PR #51, opened 2022-07-29
Fix output path

bittremieux:
Correctly use `None` for an unspecified output path by default, otherwise the previous change on lines 138-139 doesn't work.

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -81,7 +81,7 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
         tb_summarywriter=None,
         warmup_iters=100000,
         max_iters=600000,
-        output_path="",
+        output_path=None,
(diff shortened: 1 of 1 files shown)


VarunAnanth2003 (approved):
praise: Thanks for catching this!

VarunAnanth2003:
comment (nitpick): Looking at the code and where the Spec2Pep model is initialized, this isn't completely necessary as there is no place that the Spec2Pep model is initialized that **also** would output a casanovo result. (denovo mode always sets output_path=the config output path and the others don't use it). Let me know if my analysis is incorrect.

bittremieux:
The previous default of an empty string gave issues when trying to run Casanovo with the following error on the previous `makedirs` change:
```
FileNotFoundError: [Errno 2] No such file or directory: ''
```
This makes the code run again.

PR #52, opened 2022-07-29
Precursor mass filter

### LINKED ISSUE #45: Casanovo should implement the precursor filter
We should implement the precursor mass filter in Casanovo by subtracting 1 from the peptide-level score if the peptide mass is outside of a user-specified range.  This range should be specified in the yaml file in units of ppm. We can name it `precursor_window` and have  a default value of 50.

bittremieux:
WIP to modify the score if the prediction violates the precursor mass filter. Fixes #45.

Some initial stubs where calculations can be performed, but I'm not sure yet how I can retrieve the original precursor m/z's during predicting. @melihyilmaz any suggestions?


casanovo/denovo/model.py
```diff
@@ -457,39 +457,44 @@ def on_test_epoch_end(self):
...
-                        )  # select index of the last AA
-
-                        if (
-                            last_index >= 1
-                        ):  # check if peptide is at least one AA long
-                            top_scores_list = top_scores[
-                                :last_index
-                            ].tolist()  # omit the stop token
+                    peptide_seq = batch[1][i][1:]
+                    # TODO: Compute the predicted mass from the peptide sequence.
```
wsnoble, line 461: In case it's helpful, here's code to do this (copied from elsewhere):

```
AA_MASSES = {
    'A':71.03711,
    'C':103.00919,
    'D':115.02694,
    'E':129.04259,
    'F':147.06841,
    'G':57.02146,
    'H':137.05891,
    'I':113.08406,
    'J':113.08406,
    'K':128.09496,
    'L':113.08406,
    'M':131.04049,
    'N':114.04293,
    'O':114.07931,
    'P':97.05276,
    'Q':128.05858,
    'R':156.10111,
    'S':87.03203,
    'T':101.04768,
    'U':150.04344,
    'V':99.06841,
    'W':186.07931,
    'X':0, # N.B. Missing amino acid!
    'Y':163.06333,
}

MASS_WATER = 18.010564683

def compute_molecular_weight(sequence):
    """Compute the molecular weight of a protein."""
    global MASS_WATER, AA_MASSES
    
    return_value = MASS_WATER
    
    for letter in sequence:
        return_value += AA_MASSES[letter]

    return return_value
```

bittremieux: Pyteomics and spectrum_utils have functions to do that as well, I just didn't get around to it yet yesterday.

The open question is where to get the precursor m/z from, but it should be possible. I'll have to try to understand the code a bit more.

wfondrie: Depthcharge also has a calculator too!

bittremieux: I'm using depthcharge to avoid adding extra dependencies, but I think it doesn't handle the modifications properly. It does have the MassIVE-KB mods in its mass dictionary, but [this line](https://github.com/wfondrie/depthcharge/blob/5246ef8b6414855918e0823836d2432c21db63ac/depthcharge/masses.py#L78) evaluates each character individually, rather then the modified residues as an atomic unit. Can you have a look at it when you have some time @wfondrie?

wfondrie: Ah yes, I should improve the doc string there. When used I think the models, the sequence is already a list. I should add that here too.

bittremieux: Ok, good if that's the performance as intended. I fixed it on my side then.

bittremieux:
Ok, I managed to implement it.

One minor thing is that the precursor mass filter is on the _neutral_ mass of the peptides, rather than the precursor m/z. Casanovo had the former information available for the peptides, so that conversion must have happened somewhere already, although I didn't try to find out where, so I just used that rather than converting back to precursor m/z. The tolerance in ppm is going to be slightly different than when computed on precursor m/z though, so I don't know how much that would confuse people.

casanovo/denovo/model.py
```diff
@@ -77,6 +84,7 @@ def __init__(
         max_length=100,
         residues="canonical",
         max_charge=5,
+        precursor_tol_mass=50,
```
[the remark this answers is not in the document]
bittremieux: I agree. I used `precursor_tol_mass` in accordance to the `fragment_tol_mass` parameter that already existed. I'll change to `precursor_mass_tol` here, but we should probably refactor the latter separately as well then.

bittremieux:
Thanks for the review @wfondrie. I addressed your comments, except for handling the peptide strings in depthcharge. Can you ping me when that functionality is available?

wfondrie:
I should be able to get it done this afternoon. I'll let you know @bittremieux!

casanovo/denovo/model.py
```diff
@@ -200,9 +208,9 @@ def greedy_decode(self, spectra, precursors):
             a 2-tuple specifying the m/z-intensity pair for each peak. These
             should be zero-padded, such that all of the spectra in the batch
             are the same length.
-        precursors : torch.Tensor of size (n_spectra, 2)
-            The measured precursor mass (axis 0) and charge (axis 1) of each
-            tandem mass spectrum.
+        precursors : torch.Tensor of size (n_spectra, 3)
+            The measured precursor mass (axis 0), precursor charge (axis 1),
+             and precursor m/z (axis 3) of each MS/MS spectrum.
```
wfondrie, line 213: **typo:** `axis 3` -> `axis 2`

casanovo/denovo/model.py
```diff
@@ -259,9 +267,9 @@ def _step(self, spectra, precursors, sequences):
...
```
wfondrie, line 272: **typo:** `axis 3` -> `axis 2`

wfondrie (changes requested):
Looks pretty good to me. I noted a couple typos. The only other question I have is what is `casanovo_test.hdf5` used for?

bittremieux:
Ugh, the hdf5 file must have incorrectly been added to a commit. Fixed now.

wfondrie (approved):

wfondrie:
Feel free to merge @bittremieux

Issue #41, opened 2022-07-25
No mudule named pytest


yingluoqiao:
![casanovo_issue](https://user-images.githubusercontent.com/103443560/180752915-65625948-f40e-48cf-86f9-342d5c78d778.png)

bittremieux:
You can fix this error by running `pip install pytest` in your Casanovo conda environment.

However, somehow not all dependencies seem to have been installed alongside Casanovo, which is weird. It seems like you followed the installation instructions in the README, or did you deviate from them somewhere?

Alternatively, can you uninstall and reinstall Casanovo using `pip uninstall casanovo && pip install git+https://github.com/Noble-Lab/casanovo.git#egg=casanovo` and share the output here?

yingluoqiao:
Hi
 Thanks for your reply. I followed the instruction in the READMR fully. I will try out!

Best
Weiwei

获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: Wout Bittremieux ***@***.***>
发送时间: 星期一, 七月 25, 2022 5:54 下午
收件人: Noble-Lab/casanovo ***@***.***>
抄送: Peng, W. (Weiwei) ***@***.***>; Author ***@***.***>
主题: Re: [Noble-Lab/casanovo] No mudule named pytest (Issue #41)

你通常不会收到来自 ***@***.*** 的电子邮件。了解这一点为什么很重要<https://aka.ms/LearnAboutSenderIdentification>

You can fix this error by running pip install pytest in your Casanovo conda environment.

However, somehow not all dependencies seem to have been installed alongside Casanovo, which is weird. It seems like you followed the installation instructions in the README, or did you deviate from them somewhere?

Alternatively, can you uninstall and reinstall Casanovo using pip uninstall casanovo && pip install git+https://github.com/Noble-Lab/casanovo.git#egg=casanovo and share the output here?

―

bittremieux:
Fixed in #55.

Issue #38, opened 2022-05-31
IndexError on particularly ugly spectrum


kesinger:
With --mode=denovo and pretrained human weights  this spectrum:

BEGIN IONS
TITLE=crasher
RTINSECONDS=2800
PEPMASS=1551
CHARGE=4+
1552.40144197  301.23929755
1551.20606409 1051.24587242
1550.88090936 1053.2816881
END IONS

raises an IndexError at https://github.com/Noble-Lab/casanovo/blob/main/casanovo/data/datasets.py#L185

To trigger, the peaks need to be sufficiently close to each  other and to PEPMASS.

Issue #35, opened 2022-04-30
Solve pytorch ligtning warning flags and compatibility with version>= 1.6


melihyilmaz:

bittremieux:
Fixed in #55.

Issue #16, opened 2022-03-22
Training fails upon empty spectrum


guhanrv:
Hello again,

For some reason, the `on_train_epoch_end` hook is called before training is complete. I am trying to train with a pretty large dataset (~1.2M spectra between the training and validation sets, with 1093774 spectra (17,091 batches of 64) for training and 123930 spectra (1,937 batches of 64) for validation). For some reason, casanovo is thinking that training is done after only 11,927 batches (which is significantly less than the 17,091 expected) and so calls `on_train_epoch_end`, which looks for a history to add the train loss to and fails:

```
...
/home/ubuntu/smsnet_val_data/1049.mgf: 46spectra [00:00, 2628.76spectra/s]
/home/ubuntu/smsnet_val_data/1276.mgf: 62spectra [00:00, 2859.35spectra/s]
/home/ubuntu/smsnet_val_data/1073.mgf: 1spectra [00:00, 4156.89spectra/s]
/home/ubuntu/smsnet_val_data/1224.mgf: 4spectra [00:00, 5257.67spectra/s]
GPU available: True, used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
Global seed set to 454
initializing distributed: GLOBAL_RANK: 0, MEMBER: 1/1
INFO: Added key: store_based_barrier_key:1 to store for rank: 0
INFO: Rank 0: Completed store-based barrier for key:store_based_barrier_key:1 with 1 nodes.
----------------------------------------------------------------------------------------------------
distributed_backend=nccl
All distributed processes registered. Starting with 1 processes
----------------------------------------------------------------------------------------------------

LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]

  | Name    | Type             | Params
---------------------------------------------
0 | encoder | SpectrumEncoder  | 18.9 M
1 | decoder | PeptideDecoder   | 28.4 M
2 | softmax | Softmax          | 0     
3 | celoss  | CrossEntropyLoss | 0     
---------------------------------------------
47.3 M    Trainable params
0         Non-trainable params
47.3 M    Total params
189.387   Total estimated model params size (MB)
/home/ubuntu/mambaforge/envs/denovo/lib/python3.9/site-packages/pytorch_lightning/callbacks/model_checkpoint.py:631: UserWarning: Checkpoint directory /home/ubuntu/checkpoints exists and is not empty.
[...]

PR #55, opened 2022-08-04
Code refactoring

### LINKED ISSUE #54: Output file naming
I think we should add a command line parameter such as --output-root <name> that specifies the root of the output files produced by Casanovo.  Then we should have two files, <root>.casanovo.txt and <root>.casanovo.log.txt, where the former contains the PSMs and the latter contains all of the messages sent to stderr.

bittremieux:
Refactoring for readability improvements, performance increase, and new features.

Fixes #54, #35, #38, #41, #16.


casanovo/denovo/model_runner.py
```diff
@@ -0,0 +1,298 @@
...
+    Returns
+    -------
+    List[str]
+        The peak file names from the given directory.
+    """
+    return [
+        os.path.join(peak_dir, f)
+        for f in os.listdir(peak_dir)
+        if f.lower().endswith(".mgf")
+    ]
```
wfondrie, line 298: **question:** Why not use `pathlib.Path.glob()` or `glob.glob()` here?

setup.py
```diff
@@ -1,4 +1,4 @@
-"""Setup ppx"""
```
wfondrie, line 1: 😆

wfondrie (approved):
Nicely done @bittremieux. I have a few non-blocking questions, but feel free to merge when you're ready.

casanovo/denovo/model_runner.py
bittremieux: Yes, makes sense. #62 does shell parameter extension to process multiple files rather than specifying input _directories_ (which precludes that training and testing are in the same directory, which was quite annoying imo), so that PR kinda does it how you suggest.

casanovo/denovo/evaluate.py
```diff
@@ -1,347 +1,310 @@
...
-        orig = re.split(r"(?<=.)(?=[A-Z])", true_pep_seqs[pred_ind])
-        orig_total_num_aa += len(orig)
-        pred_total_num_aa += len(pred)
-
-        aa_match, pep_match = match_aa(
-            orig, pred, aa_dict, eval_direction=eval_direction
+    aa_matches_batch, n_aa1, n_aa2 = [], 0, 0
+    for peptide1, peptide2 in zip(peptides1, peptides2):
+        tokens1 = re.split(r"(?<=.)(?=[A-Z])", peptide1)
+        tokens2 = re.split(r"(?<=.)(?=[A-Z])", peptide2)
```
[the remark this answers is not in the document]
bittremieux: I agree. I'll create an issue in depthcharge to track this, and we can replace these two lines later then.

Issue #21, opened 2022-04-10
Add delta mass to model output


melihyilmaz:

bittremieux:
Fixed in #59.

Issue #27, opened 2022-04-11
Document input and output


melihyilmaz:
On [README.md](https://github.com/Noble-Lab/casanovo/blob/main/README.md) and on the website.

melihyilmaz:
We have  
> Casanovo can predict peptide sequences for MS/MS spectra in mzML, mzXML, and MGF files. This will write peptide predictions for the given MS/MS spectra to the specified output file in mzTab format

in the `Getting Started` page

bittremieux:
Yes, but I think @wsnoble also wanted a brief description of what the output mzTab file looks like. So explain what information is included in the metadata and PSM sections.

wsnoble:
Yes.  I think it would be helpful to explain what the various columns mean.  If possible, please also include pointers to descriptions of the mzML, mzXML, MGF and mzTab format specifications.

PR #59, opened 2022-08-05
Export to mzTab

### LINKED ISSUE #20: Provide model output in mzTab format

bittremieux:
Fixes #20 and #21.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ See [`casanovo/config.yaml`](https://github.com/Noble-Lab/casanovo/blob/main/cas
 casanovo --mode=denovo --model='path/to/pretrained.ckpt' --peak_dir='path/to/predict/mgf/files/dir' --config='path/to/config.yaml' --output='path/to/output'
 ```
 
-This will write peptide predictions for the given MS/MS spectra to the specified output file in a tab-separated format (extension: .csv).
+This will write peptide predictions for the given MS/MS spectra to the specified output file in mzTab format.
 
 - To evaluate _de novo_ sequencing performance based on known spectrum annotations:
 
@@ -101,17 +101,9 @@ The example MGF file is available at [`sample_data/sample_preprocessed_spectra.m
 casanovo --mode=denovo --model='[PATH_TO]/pretrained_excl_mouse.ckpt' --peak_dir='sample_data' --config='path/to/config.yaml'
 ```
 Make sure you use the proper filepath to the `pretrained_excl_mouse.ckpt` file.
-    - Note: If you want to get the ouput CSV file in different location than the working directory, specify an alternative output location using the `--output` parameter.
+    - Note: If you want to get the output mzTab file in different location than the working directory, specify an alternative output location using the `--output` parameter.
 
 This job will take very little time to run (< 1 minute).
-
-If the first few lines look like:
-
-```
-spectrum_id     sequence            score       aa_scores
-0               LAHYNKR             0.98197     1.0,0.999...
-1               VKEDYGQM+15.995PR   0.77206     0.999,0.999...
-```
 
 Congratulations! You got Casanovo to work.
 
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -226,7 +226,7 @@ def _write_mztab_header(
             "psm_search_engine_score[1]",
             "[MS, MS:1001143, search engine specific score for PSMs, ]",
         ),
-        ("software[1]", f"[MS, MS:1001456, Casanovo, {__version__}]"),
+        ("software[1]", f"[MS, MS:1003281, Casanovo, {__version__}]"),
     ]
     if len(fixed_mods) == 0:
         metadata.append(
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -7,6 +7,7 @@ import os
 import pathlib
 import re
 import sys
+from typing import Any, Dict
 
 import click
 import pytorch_lightning as pl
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -101,18 +101,10 @@ The example MGF file is available at [`sample_data/sample_preprocessed_spectra.m
 casanovo --mode=denovo --model='[PATH_TO]/pretrained_excl_mouse.ckpt' --peak_dir='sample_data' --config='path/to/config.yaml'
 ```
 Make sure you use the proper filepath to the `pretrained_excl_mouse.ckpt` file.
-    - Note: If you want to get the ouput CSV file in different location than the working directory, specify an alternative output location using the `--output` parameter.
+    - Note: If you want to get the output mzTab file in different location than the working directory, specify an alternative output location using the `--output` parameter.
 
 This job will take very little time to run (< 1 minute).
```
wfondrie, line 107: **question (non-blocking):** Would it be good to provide a few example lines from the mzTab file?

casanovo/casanovo.py
```diff
@@ -1,8 +1,13 @@
 """The command line entry point for Casanovo."""
+import collections
+import csv
 import datetime
 import logging
 import os
+import pathlib
```
wfondrie, line 7: **question:** Typically, pathlib is used as `from pathlib import Path`. Is there a particular reason to break that convention here? [docs](https://docs.python.org/3/library/pathlib.html#basic-use)

casanovo/casanovo.py
```diff
@@ -130,13 +138,26 @@ def main(
 
     # Log the active configuration.
     logger.info("Casanovo version %s", str(__version__))
+    logger.debug("mode = %s", mode)
+    logger.debug("model = %s", model)
+    logger.debug("peak_dir = %s", peak_dir)
+    logger.debug("peak_dir_val = %s", peak_dir_val)
+    logger.debug("config = %s", config_fn)
+    logger.debug("output = %s", output)
```
wfondrie, line 146: **praise:** This is a nice addition that will help us with troubleshooting folks' issues!

casanovo/casanovo.py
```diff
@@ -130,13 +138,26 @@ def main(
...
     if mode == "denovo":
         logger.info("Predict peptide sequences with Casanovo.")
-        model_runner.predict(peak_dir, model, output, config)
+        _write_mztab_header(
+            f"{output}.mztab",
+            peak_dir,
+            config,
+            model=model,
+            config_filename=config_fn,
+        )
```
wfondrie, line 159: **question:** Do you think we should clean up the file in the case that the predict step fails?

casanovo/denovo/model.py
```diff
@@ -12,6 +12,7 @@
 from depthcharge.models.embed.model import PairedSpectrumEncoder
```
wfondrie, line 12: **question (non-blocking, irrelevant):** Nothing to do with this PR, but do you know why this import exists here?

casanovo/casanovo.py
```diff
@@ -145,5 +166,112 @@ def main(
...
+            metadata.append(
+                (f"variable_mod[{i}]-site", aa if aa else "N-term")
+            )
+    for i, (key, value) in enumerate(kwargs.items(), 1):
+        metadata.append((f"software[1]-setting[{i}]", f"{key} = {value}"))
+    for i, (key, value) in enumerate(config.items(), len(kwargs) + 1):
+        if key not in ("residues",):
+            metadata.append((f"software[1]-setting[{i}]", f"{key} = {value}"))
+    with open(filename_out, "w") as f_out:
+        writer = csv.writer(f_out, delimiter="\t")
```
wfondrie, line 271: **suggestion:** I think I agree with @wsnoble that we should use OS-specific line endings. What do you think?

It should be easy enough:
```python
writer = csv.writer(f_out, delimiter="\t", lineterminator=os.linesep())
```

casanovo/denovo/model.py
```diff
@@ -404,9 +402,32 @@ def on_predict_epoch_end(
         if self.out_filename is None:
             return
         empty_token_score = torch.tensor(0.04)
-        with open(self.out_filename, "w") as f_out:
+        with open(self.out_filename, "a") as f_out:
             writer = csv.writer(f_out, delimiter="\t")
```
wfondrie, line 406: **suggestion:** Similar to above, I think we should use OS-specific line endings:

```python
writer = csv.writer(f_out, delimiter="\t", lineterminator=os.linesep())
```

wfondrie (changes requested):
Looks great - I just had a couple minor questions and comments.

README.md
bittremieux: It will probably be useful to explain the format to some extent, but because there's a metadata section and then the PSM section, it's kind of verbose to include it in the README. I'd suggest to add a proper description when we set up a full documentation website on readthedocs. We already have #27 to remind us of it.

casanovo/denovo/model.py
bittremieux: It has always been allowed as an option for the encoder of a `Spec2Pep` object, although we always just use a `SpectrumEncoder` for Casanovo. Maybe a dangling copy/paste issue?

casanovo/casanovo.py
bittremieux: **praise:** Excellent solution.

casanovo/casanovo.py
bittremieux: Good suggestion to avoid confusion with incomplete files. I surrounded the predicted with a `try ... except` where the `expect` doesn't catch a specific error, which can be frowned upon. I think in this case it's probably warranted to have such a general `except` clause though. What do you think?

bittremieux: Solved by refactoring the mzTab export functionality.

casanovo/denovo/model.py
wfondrie: I think so. We can delete in another PR.

wfondrie (approved):
Looks good to me. Feel free to merge.

PR #60, opened 2022-08-05
Update residues alphabet

bittremieux:
Because our main model will be trained on MassIVE-KB, this adds the corresponding modifications to our residues alphabet. Note that the updated alphabet **can't** be used with our currently trained models.

To ensure that residues are always interpreted as strings by the YAML loader, they are now surrounded by quote marks. Otherwise N-terminal modifications are interpreted as floats.


VarunAnanth2003:
**nitpick/non-blocking:** Maybe it would be helpful to mark which amino acids we needed to add because of the MSKB training job with a comment? Not sure if people would look at the paper and be confused as to why certain AA seem to have been added later. If this is not needed then this seems fine since it follows the old format for the residues section of the yaml.

bittremieux:
Imo this is more a problem of the fact that we didn't do versioning with proper releases previously. At this point there are a few changes that are not fully backwards compatible with the code for the ICML paper, so just adding a comment for the residues won't be a complete solution anyway.

I'd just try to get the new model available asap and make a real release, so it's much easier and clearer for people what proper version to use.

VarunAnanth2003 (approved):
Forgot to do this earlier - this looks good and consistent with how the .yaml should be formatted

PR #62, opened 2022-08-06
Specify peak files as shell pattern

bittremieux:
Specify peak files as a shell pattern rather than a directory. This makes it possible to run different Casanovo tasks on files in the same directory, rather than having to create unique directories per peak file.

Example usage:

- One file: `--peak_path=~/casanovo/spectra.mgf`
- All files in directory: `--peak_path=~/casanovo/*.mgf`


casanovo/denovo/model_runner.py
```diff
@@ -278,22 +265,25 @@ def train(
...
-        The peak file names from the given directory.
+        The peak file names matching the path pattern.
     """
     return [
-        os.path.join(peak_dir, f)
-        for f in os.listdir(peak_dir)
-        if f.lower().endswith(".mgf")
+        fn
+        for fn in click.utils._expand_args([path])
+        if os.path.splitext(fn.lower())[1] == ".mgf"
```
wfondrie, line 288: **comment:** I think we should avoid using a private function in click to do the work here.

Maybe we should add our own utility function here? I could be swayed either way.

wfondrie (approved):
This looks good to me. I left a couple comments/questions to consider though. Feel free to merge or update as you see fit @bittremieux 👍

casanovo/denovo/model_runner.py
```diff
@@ -105,15 +106,9 @@ def _execute_existing(
...
-    if not os.path.isdir(peak_dir):
-        logger.error(
-            "Could not find directory %s from which to read peak files",
-            peak_dir,
-        )
-        raise FileNotFoundError(
-            "Could not find the directory to read peak files"
-        )
-    peak_filenames = _get_peak_filenames(peak_dir)
+    if len(peak_filenames := _get_peak_filenames(peak_path)) == 0:
```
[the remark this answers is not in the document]
bittremieux: I'm not an extensive user of the walrus operator either. In long if-else statements it can be useful to make code more concise and help readability, although here it only saved us a single line.

casanovo/denovo/model_runner.py
bittremieux: I mainly wanted to get the glob functionality, but after reading some [discussions on click's shell functionality](https://github.com/pallets/click/issues/1096) I found this function that gives user and variable expansion for free compared to using glob.

But it makes sense that we can't depend on having a consistent API for private functions. I'll implement something similar explicitly.

PR #58, opened 2022-08-05
Update config.yaml

VarunAnanth2003:
Fixes issue with certain .yaml attributes being read as strings

PR #65, opened 2022-08-08
Ensure that config values have the correct type

### LINKED ISSUE #64: Correct type of YAML config values
Make sure that the YAML config values are parsed as the correct type. I.e. learning rate `5e-4` should be a float, not a string.

bittremieux:
Fixes floats being interpreted as strings and other type mismatches. Incorrect types will give an error with corresponding error message:
> 2022-08-08 08:32:18,799 ERROR [casanovo/MainProcess] casanovo.main : Incorrect type for configuration value learning_rate: could not convert string to float: 'bla'

Fixes #64, supersedes #58.


VarunAnanth2003:
**Issue/question**: How will this code format handle the tb_summarywriter field? In the code, it needs to be passed in as a SummaryWriter object and not None to allow Tensorboard to work. I am not sure how we can handle this in yaml format. **Perhaps rewrite code to handle writing to tensorboard as a boolean?**

bittremieux:
Because `tb_summarywriter` is not one of the keys to check (only the types of int, float, and boolean keys are validated), nothing different will happen with it than previously.

That said, I haven't used Tensorboard here yet, so I must admit I'm not sure how this config option is supposed to work from the YAML file. Because the code indeed expects a SummaryWriter object, whereas you can't specify that in YAML. Did it even work properly previously?

casanovo/casanovo.py
```diff
@@ -121,10 +121,62 @@ def main(
...
+            except (TypeError, ValueError) as e:
+                logger.error(
+                    "Incorrect type for configuration value %s: %s", key, e
+                )
+                raise TypeError(
+                    f"Incorrect type for configuration value {key}: {e}"
+                )
+    config["residues"] = {
+        str(aa): float(mass) for aa, mass in config["residues"].items()
+    }
```
wfondrie, line 179: **suggestion:** This works, but I think it might be easier to maintain if we had a dictionary that mapped each parameter to its expected type instead.

wfondrie (approved):
Nice addition. I made one, non-blocking suggestion.

PR #66, opened 2022-08-08
Train optimization

bittremieux:
Minor changes to hopefully train a bit faster on large datasets. At the very least it should use the current number of CPU cores that are available and avoid some warning messages.


wfondrie (approved):
Looks good to me 👍

PR #67, opened 2022-08-20
Fix deleted config values

bittremieux:
I forgot to harmonize changes across PRs for these two config values.


VarunAnanth2003 (approved):
approval: Looked at line 278 of model_runner.py and this name change should keep the naming consistent.

PR #68, opened 2022-08-23
Automatically download model weights

### LINKED ISSUE #17: Casanovo should automatically download the pre-trained model weights and be able to use them easily.

### LINKED ISSUE #75: Errror running example job
When running the provided  following command:
casanovo --mode=denovo --model=22-07-02_weights/pretrained_excl_mouse.ckpt --peak_path=sample_data/sample_preprocessed_spectra.mgf --config=casanovo/config.yaml

This error shows:
 File "D:\miniconda3\lib\site-packages\torch\nn\modules\module.py", line 1497, in load_state_dict
    raise RuntimeError('Error(s) in loading state_dict for {}:\n\t{}'.format(
RuntimeError: Error(s) in loading state_dict for Spec2Pep:
        size mismatch for decoder.aa_encoder.weight: copying a param with shape torch.Size([25, 512]) from checkpoint, the shape in current model is torch.Size([29, 512]).
        size mismatch for decoder.final.weight: copying a param with shape torch.Size([25, 512]) from checkpoint, the shape in current model is torch.Size([29, 512]).
        size mismatch for decoder.final.bias: copying a param with shape torch.Size([25]) from checkpoint, the shape in current model is torch.Size([29]).

Did you recently updated depthcharge-ms, the pretrained weights, the sample .mgf file or config.yaml?
Shall I try any older version of any of these files? Thanks!

bittremieux:
Fixes #17. Fixes #75.

Use cached model weights or download them from GitHub.

If no weights file (extension: .ckpt) is available in the cache directory, it will be downloaded from a release asset on GitHub. Model weights are retrieved by matching release version. If no model weights for an identical release (major, minor, patch), alternative releases with matching (i) major and minor, or (ii) major versions will be used. If no matching release can be found, no model weights will be downloaded.

Note that the GitHub API is limited to 60 requests from the same IP per hour. A log message provides instructions to explicitly specify the model file for subsequent uses.

Review: @wsnoble to check whether this is the desired behavior for the users and the documentation, @wfondrie to check the code.


casanovo/casanovo.py
```diff
@@ -179,5 +207,105 @@ def main(
...
+        The name of the model weights file.
+    """
+    cache_dir = appdirs.user_cache_dir("casanovo", False, opinion=False)
+    os.makedirs(cache_dir, exist_ok=True)
+    version = tuple(
+        g
+        for g in re.match(
+            r"(\d+)\.(\d+)\.(\d+)(?:.dev\d+.+)?", __version__
+        ).groups()
+    )
```
wfondrie, line 236: **suggestion:** This is pretty ugly. Maybe use `re.compile` above to clean it up? Something like:
```python
version_regex = re.compile(r"(\d+)\.(\d+)\.(\d+)(?:.dev\d+.+)?")
version = tuple(g for g in version_regex.match(__version__).groups())
```

wfondrie (changes requested):
This looks like it solves the problem to me. Its a bit complicated and I think it'd be worthwhile to create unit tests for it, perhaps by mocking a response GitHub.

Also, do we have a way to verify that the architecture (number of layers, layer dim, etc) matches the downloaded weights, or to automatically set it?

wfondrie:
It looks like checking the version is failing in GH Actions 🤔

bittremieux:
I see you removed the `psutil` dependency. This is necessary to correctly set resources on a shared system, such as the cluster. `os.cpu_count()` will give the number of CPUs that are present, `psutil` instead will give the number CPUs that can be used. Thus, if you use `os.cpu_count()` but you requested only a specific number of CPUs, this will make sure that there aren't an excessive number of worker threads (extreme case: on n032 use 40 workers threads when the job only has 1 CPU).

wfondrie:
Good point @bittremieux, but `psutils.Process.cpu_affinity()` doesn't support MacOS. We need another solution 🤔

wfondrie:
@bittremieux, would `len(os.sched_getaffinity())` work? https://docs.python.org/3/library/os.html#os.sched_getaffinity

wfondrie:
Sounds like we may need something like:
```python
import os
import psutils

def n_workers():
    try:
        return len(psutils.Process().cpu_affinity())
    except AttributeError:
        return len(os.sched_getaffinity(0))
```
Or vice versa. Thoughts?

bittremieux:
Yes, that makes sense. I was unaware of the Mac incompatibility because I don't have a Mac. Likewise, whether it runs on Windows is only from reading StackOverflow and the psutils documentation, not from exactly testing it.

wfondrie:
Now we're getting a multiprocessing error on Windows 🤦‍♂️

wfondrie (approved):

PR #69, opened 2022-08-24
Simplify config

bittremieux:
1. Harmonize default parameters in the code and in the config file, as per the discussion with Melih on Slack (Aug 6).
2. Simplify the examples to not use a config file. As sensible default values are used, we can simplify how most users will interact with Casanovo.


VarunAnanth2003 (approved):
approval: Looks good!

PR #71, opened 2022-08-29
Account for incorrect AAs during evaluation

bittremieux:
Don't crash when a prediction that's not found in the AA alphabet is encountered.


VarunAnanth2003 (approved):
**non-blocking question**: How does switching from array access to .get() account for incorrect AA's? Does it just prevent an error from being thrown?

bittremieux:
`get` on a dictionary indeed doesn't result in a `KeyError`, but will return a default value instead if the key is missing. By default this is `None`, but here I have it return a mass of `0` instead. This will cause the AA masses to not match without an error occurring.

PR #72, opened 2022-08-30
Account for non-monoisotopic precursors

bittremieux:
Don't penalize peptide scores if the precursor m/z fits a C13 difference. By default, a single C13 shift is allowed (this can be changed if we want to be less strict).

I verified it with a few spectra where the "ground truth" were deamidated peptides and Casanovo didn't predict a modification but got negative scores because the precursor m/z tolerance was violated. With this change, the scores are positive instead for these correct identifications.

I also added the `precursor_mass_tol` option to the config file, which was apparently still missing.


casanovo/casanovo.py
```diff
@@ -126,14 +126,16 @@ def main(
         max_mz=float,
         min_intensity=float,
         remove_precursor_tol=float,
+        max_charge=int,
+        precursor_mass_tol=float,
+        isotope_error_range=lambda min_max: (int(min_max[0]), int(min_max[1])),
```
wfondrie, line 131: **question:** Is there a reason to use a range here instead of just specifying the maximum isotope error?

casanovo/denovo/model.py
```diff
@@ -420,11 +426,24 @@ def on_predict_epoch_end(
...
+                                    - isotope * 1.00335 / precursor_charge
+                                )
+                            )
+                            / precursor_mz
+                            * 10**6
+                            for isotope in range(
+                                self.isotope_error_range[0],
+                                self.isotope_error_range[1] + 1,
+                            )
+                        ]
```
wfondrie, line 443: **suggestion:** This is pretty difficult to read. What about adding a function:
```python
def calc_mass_error(calc_mz, obs_mz, isotope, charge):
    return (calc_mz - (obs_mz - isotope * 1.00335) / charge)) / obs_mz * 1e6
```

Then we could simplify this section to:
```python
delta_mass_ppm = [
    calc_mass_error(calc_mz, precursor_mz, precursor_charge, i) 
    for i in range(self.isotope_error_range[0], self.isotope_error_range[1] + 1)
]
```

wfondrie (changes requested):
Nice! I had one question and one suggestion for clarity.

casanovo/casanovo.py
bittremieux: The range makes it possible to not only look at a previous isotopic peak (subtract C13 mass difference), but also at a next isotopic peak (add C13 mass difference). Most of the time this is probably not necessary, but for some molecules it can be that the mono-isotopic peak is in the middle of the isotopologue series. See e.g. [here](https://pubs.acs.org/doi/10.1021/jasms.1c00110).

It's an edge case though, so maybe this is overkill. Should we simplify it to only look backwards?

wfondrie: I think it's fine to leave it how it is. Thanks!

wfondrie (approved):

wfondrie:
Feel free to merge when you're ready 👍

PR #79, opened 2022-10-03
Update for PyPI release

wfondrie:
This PR sets up everything for the PyPI release. New GitHub releases will automatically uploaded to PyPI.


README.md
```diff
@@ -52,7 +52,7 @@ The base environment most likely will not work.
 Install Casanovo as a Python package from this repository (requires 3.8 ≥ [Python version] > 3.10 , dependencies will be installed automatically as needed):
```
bittremieux, line 52: **question:** Does Python (fully) 3.10 work? The instructions currently dictate Python 3.8 or 3.9, whereas the tests Action uses Python 3.10.

setup.cfg
```diff
@@ -22,7 +22,7 @@ packages = find:
 python_requires = >=3.8
 install_requires =
     click
-    depthcharge-ms @ git+https://github.com/wfondrie/depthcharge@3dda29053ab479fa31fa1309f65f285d04481c1e
+    depthcharge-ms==0.0.1
```
bittremieux, line 25: **suggestion:** I'd use `>=` to get the latest optional depthcharge improvements without having to update the Casanovo dependencies every time.

bittremieux:
PR #68 also includes changes to make the tests working again, so we have to be careful when integrating the changes from both PRs.

setup.cfg
wfondrie: Whoops! Good catch 😅

wfondrie:
Thanks @bittremieux for the heads up. What do you think we should do for now? 

I see three options:
1. Migrate the working unit test changes to this PR and merge.
2. Wait to merge this branch until #68 is ready.
3. Merge this branch with failing tests.

I lean toward 1, but what do you think?

README.md
wfondrie: I actually don't know. Do you know why it didn't work before? My thought is we may have tried it shortly after 3.10 was released, before most packages had a chance to update.

bittremieux: I haven't tested it with 3.10 and just followed the existing description in the README, so I don't know.

casanovo/casanovo.py
```diff
@@ -163,8 +163,11 @@ def main(
     }
     # Add extra configuration options and scale by the number of GPUs.
     n_gpus = torch.cuda.device_count()
-    config["n_workers"] = len(psutil.Process().cpu_affinity()) // n_gpus
-    config["train_batch_size"] = config["train_batch_size"] // n_gpus
+    if n_gpus:
+        config["n_workers"] = len(psutil.Process().cpu_affinity()) // n_gpus
+        config["train_batch_size"] = config["train_batch_size"] // n_gpus
+    else:
+        config["n_workers"] = 0
```
wfondrie, line 170: @bittremieux - I got a div by zero error when running only on CPU. Do you think using 0 workers by default on CPU makes sense? I haven't really looked how number of workers affects inference on CPU. Maybe 1 is better?

bittremieux: Rather than 0 or 1, for optimal use of the system resources I'd set to `len(psutil.Process().cpu_affinity())` instead. Dividing by the number of GPUs is only because Pytorch Lightning starts multiple independent instances of the code when running on multiple GPUs.

So I'd change these lines like this:
```
config["n_workers"] = len(psutil.Process().cpu_affinity())
if n_gpus > 1:
    config["n_workers"] = config["n_workers"] // n_gpus
    config["train_batch_size"] = config["train_batch_size"] // n_gpus
```
That should work for both CPU and (multiple) GPUs.

README.md
wfondrie: I haven't run into trouble with 3.10 yet, but maybe we should wait to change the README until more tests are added.

bittremieux (approved):

PR #81, opened 2022-10-09
Prevents divide by zero error caused when n_gpus == 0

VarunAnanth2003:
Error was on line 166 of casanovo.py (```config["n_workers"] = len(psutil.Process().cpu_affinity()) // n_gpus```) and caused the error: ```ZeroDivisionError: integer division or modulo by zero```. Added a safety net if ```n_gpus```  is zero.


bittremieux:
This is already being discussed in #79. To prevent merge conflicts, I'll close this PR.

PR #85, opened 2022-10-23
Adding .csv generation to eval mode of Casanovo

VarunAnanth2003:
If merged, will allow for the generation of an evaluation graph when running casanovo in `eval` mode. The location of the graph is specified with the `--output` CLI option.


wfondrie:
I've been working on this today and I've come to the conclusion that I don't think it's feasible to make the benchmarking more general. The challenge is converting the modification strings from one format to the same used by MassIVE-KB. The major issue with this conversion is that some sources don't include fewer decimal places in their modification strings, making the conversion non-trivial.

@VarunAnanth2003 - how goes removing the actual plots and instead returning a DataFrame? I'll be making some small changes to docstrings and such.

bittremieux:
Can you give an example of this conversion problem @wfondrie? [spectrum_utils now supports the full ProForma 2.0 specification](https://spectrum-utils.readthedocs.io/en/latest/annotating.html), which is reasonably flexible in the way modifications can be specified. IIRC the MassIVE-style modifications are not fully ProForma-compliant, but can easily be converted by surrounding the masses by square brackets. Could we try to use ProForma as an intermediate representation?

wfondrie:
That seems like good idea @bittremieux! Here one the current conversions that we're doing in this PR:
```python
    mgf_df["true_seq"] = mgf_df["true_seq"].str.replace(
        "C[57.02]", "C+57.021", regex=False
    )
```
Here, `C[57.02]` is the modification output by Tide (right @VarunAnanth2003?), and `C+57.021` is the equivalent in our MassIVE-KB vocabulary. This example is trivial to convert manually, but difficult otherwise, because we'd need some kind of fuzzy matching scheme. Is there an easy way to convert all of these to proforma?

bittremieux:
Hmm, the difference in precision makes it a bit challenging. ProForma allows to specify modifications by their mass as `PEPC[+57.021]IDER`, so slightly different than both of those representations. Converting each one to a compatible ProForma string is straightforward, but that won't take care of the different number of decimals.

I guess we could round to the number of decimals from the least precise representation? For that we don't even need ProForma, but maybe that's still useful as a consistent representation. In which case the steps would be:
1. Surround all numeric values by square brackets and always show the sign.
2. Round to a specified number of decimals (2).

Not entirely satisfactory, but it maybe works?

casanovo/denovo/evaluate.py
```diff
@@ -310,3 +315,130 @@ def aa_precision_recall(
...
+    --------
+    mgf_df : pandas DF
+        The true peptide labels with column "true_seq"
+    """
+    seqs = []
+    with mgf.MGF(mgf_filename) as f_in:
+        for spectrum_dict in f_in:
+            seqs.append(spectrum_dict["params"]["seq"])
+    mgf_df = pd.DataFrame({"true_seq": seqs})
+    mgf_df["true_seq"] = mgf_df["true_seq"].str.replace(
```
melihyilmaz, line 339: If 'C' not represented with its permanent modification, this implementation fails to match those with 'C+57.021' token predicted by Casanovo

casanovo/denovo/evaluate.py
```diff
@@ -310,3 +315,130 @@ def aa_precision_recall(
...
+    mgf_filename : str
+        Path to the .mgf file to extract from.
+
+    Returns:
+    --------
+    mgf_df : pandas DF
+        The true peptide labels with column "true_seq"
+    """
+    seqs = []
+    with mgf.MGF(mgf_filename) as f_in:
```
melihyilmaz, line 335: We can generalize this to a list of mgf files rather than a single file at a time

melihyilmaz (changes requested):
In the context of conversation above around generalized PTM representations but of more immediate practical concern, "C" is represented without its default modification currently impacts precision calculation for 9 species benchmark.

wsnoble:
We should work on merging this.  I think it will require that Varun write some tests.

melihyilmaz:
@wfondrie With Wout's proposal for generalizing how modifications are represented, do you think there's a way forward with generalized benchmarking or should we go ahead with the current treatment of modifications?

VarunAnanth2003:
I can work on writing tests/finishing this PR after figuring out how to go forward

PR #86, opened 2022-10-25
Pass folder path to initialize Tensorboard object

melihyilmaz:
Changed `tb_summarywriter` from `torch.utils.tensorboard.SummaryWriter` to `str` because we used to initialize the object in **config.py** but can't do this in **config.yaml**. Initialization now happens inside `Spec2Pep` if a folder path is passed.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,4 +40,5 @@ jobs:
     - name: Upload coverage to codecov
       uses: codecov/codecov-action@v3
       with:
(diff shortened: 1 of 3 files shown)


wfondrie (changes requested):
Looks good, but we need to add a unit test 😉.

wfondrie (approved):

wfondrie:
Feel free to merge when ready! 🎉 

Also, please make sure to use "Squash and Merge"

PR #87, opened 2022-11-02
Add beam search peptide decoding

melihyilmaz:
Add beam search decoding to replace greedy decoding. Organized `beam_search_decode()` as a series of sub-functions to allow for unit testing each.

From a larger group of peptides predictions, this implementation of beam search decoding caches _k_ highest scoring peptide predictions for each spectrum, prioritizing the peptides fitting the observed precursor mass. As an output, the highest scoring peptide within precursor m/z tolerance, is returned, i.e. a single PSM for each spectra is recorded in the output mzTab file. If there are no cached peptides within precursor m/z tolerance for a spectrum, the highest scoring peptide  is returned in the output.

Also fixed amino acid-level score calculation in `model.on_predict_epoch_end()`, which was previously retrieving a list of shifted-by-one-AA amino acid scores, i.e. first AA prediction being assigned the second score, second AA the third score etc. where last AA was assigned the score for the stop toke.

### DIFF
--- a/casanovo/config.yaml
+++ b/casanovo/config.yaml
@@ -8,7 +8,7 @@ random_seed: 454
 
 # Spectrum processing options.
 n_peaks: 150
-min_mz: 50.52564895  # 1.0005079 * 50.5
+min_mz: 50.0
 max_mz: 2500.0
 min_intensity: 0.01
 remove_precursor_tol: 2.0  # Da
@@ -21,7 +21,7 @@ dim_model: 512
 n_head: 8
 dim_feedforward: 1024
 n_layers: 9
-dropout: 0
+dropout: 0.0
 dim_intensity:
 custom_encoder:
 max_length: 100
--- a/casanovo/config.yaml
+++ b/casanovo/config.yaml
@@ -65,7 +65,7 @@ weight_decay: 1e-5
 # Training/inference options.
 train_batch_size: 32
 predict_batch_size: 1024
-n_beams: 1
+n_beams: 5
 
 logger:
 max_epochs: 30
--- a/casanovo/config.yaml
+++ b/casanovo/config.yaml
@@ -65,6 +65,7 @@ weight_decay: 1e-5
 # Training/inference options.
 train_batch_size: 32
 predict_batch_size: 1024
+n_beams: 1
 
 logger:
 max_epochs: 30
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -1,7 +1,15 @@
 """A de novo peptide sequencing model."""
 import logging
 import re
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import (
+    Any,
+    Dict,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Union,
+)
 from heapq import heappop, heappush
 
 import depthcharge.masses
@@ -191,12 +199,15 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
         )
         return [self.decoder.detokenize(t) for t in tokens], aa_scores
 
-    def beam_search_decode(self, spectra, precursors):
-        """Beam search decode the spectra.
+    def beam_search_decode(
+        self, spectra: torch.Tensor, precursors: torch.Tensor
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """
+        Beam search decode the spectra. Return the highest scoring peptide, within the precursor m/z tolerance whenever possible
 
         Parameters
         ----------
-        spectrum : torch.Tensor of shape (n_spectra, n_peaks, 2)
+        spectra : torch.Tensor of shape (n_spectra, n_peaks, 2)
             The spectra to embed. Axis 0 represents a mass spectrum, axis 1
             contains the peaks in the mass spectrum, and axis 2 is essentially
             a 2-tuple specifying the m/z-intensity pair for each peak. These
@@ -230,20 +241,14 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
         tokens = torch.zeros(batch, length, beam)
         tokens = tokens.type_as(spectra).long()
 
-        # Cache terminated beams and their scores
-        cache_scores = einops.rearrange(scores.clone(), "B L V S -> (B S) L V")
-        cache_tokens = einops.rearrange(tokens.clone(), "B L S -> (B S) L")
-
-        # Keep pointer to free rows in the cache and already cached predictions
-        cache_idx_dict = {
-            spec_idx: spec_idx * beam for spec_idx in range(batch)
-        }
-        # Keep already decoded peptides to avoid duplicates in cache
-        cache_seq_dict = {spec_idx: set() for spec_idx in range(batch)}
-        # Store peptide scores to replace lower score peptides in cache
-        # with higher score peptides during decoding.
-        cache_score_dict = {spec_idx: [[], []] for spec_idx in range(batch)}
-
+        # Create cache for decoded beams
+        (
+            cache_scores,
+            cache_tokens,
+            cache_idx_dict,
+            cache_seq_dict,
+            cache_score_dict,
+        ) = self._create_beamsearch_cache(scores, tokens)
         # Create the first prediction:
         pred, _ = self.decoder(None, precursors, memories, mem_masks)
         _, idx = torch.topk(pred[:, 0, :], beam, dim=1)
@@ -260,9 +265,12 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
             scores = einops.rearrange(scores, "B L V S -> (B S) L V")
             tokens = einops.rearrange(tokens, "B L S -> (B S) L")
 
-            finished_beams_idx = self._terminate_finished_beams(
+            # Terminate beams exceeding precursor m/z tolerance and track all terminated beams
+            finished_beams_idx, tokens = self._terminate_finished_beams(
                 tokens, precursors, idx
             )
+
+            # Cache terminated beams, group and order by fitting precursor m/z and confidence score
             (
                 cache_idx_dict,
                 cache_seq_dict,
@@ -280,9 +288,10 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
                 idx,
             )
 
-            # Terminate when all current beams are terminated
+            # Stop decoding when all current beams are terminated
             decoded = (tokens == self.stop_token).any(axis=1)
             if decoded.all():
+                # Return the peptide with highest confidence score (fitting precursor m/z if possible)
                 output_tokens, output_scores = self._get_top_peptide(
                     cache_score_dict, cache_tokens, cache_scores, batch
                 )
@@ -297,13 +306,96 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
                 mem_masks[~decoded, :],
             )
 
+            # Find top-k beams with highest confidence scores and continue decoding from those
             scores, tokens = self._get_topk_beams(scores, tokens, batch, idx)
 
         return self.softmax(output_scores), output_tokens
 
-    def _terminate_finished_beams(self, tokens, precursors, idx) -> List[int]:
-        print("tokens", tokens.shape, tokens)
-        print("precursors", precursors.shape, precursors)
+    def _create_beamsearch_cache(
+        self, scores: torch.Tensor, tokens: torch.Tensor
+    ) -> Tuple[
+        torch.Tensor,
+        torch.Tensor,
+        Dict[int, int],
+        Dict[int, Set[str]],
+        Dict[int, List[List[Tuple[float, int]]]],
+    ]:
+        """
+        Create cache tensor and dictionary to store and group terminated beams during beam search decoding
+
+        Parameters
+        ----------
+        scores : torch.Tensor of shape (n_spectra, max_length, n_amino_acids, n_beams)
+            Output scores of the model.
+        tokens : torch.Tensor of size (n_spectra, max_length, n_beams)
+            Output token of the model corresponding to amino acid sequences of peptides.
+
+        Returns
+        -------
+        cache_scores : torch.Tensor of shape
+        (n_spectra* n_beams, max_length, n_amino_acids)
+            The score for each amino acid in cached peptides.
+        cache_tokens : torch.Tensor of shape
+        (n_spectra* n_beams, max_length)
+            The token for each amino acid in cached peptides.
+        cache_idx_dict : Dict[int, int]
+            Next available tensor index to cache peptides for each spectrum.
+        cache_seq_dict : Dict[int, Set[str]]
+            Set of decoded peptides for each spectrum.
+        cache_score_dict : Dict[int, List[List[Tuple[float, int]]]
+            Confidence score for each decoded peptide, separated as precursor m/z fitting vs not, for each spectrum.
+
+        """
+
+        batch = scores.shape[0]
+        beam = scores.shape[-1]
+
+        # Cache terminated beams and their scores
+        cache_scores = einops.rearrange(scores.clone(), "B L V S -> (B S) L V")
+        cache_tokens = einops.rearrange(tokens.clone(), "B L S -> (B S) L")
+
+        # Keep pointer to free rows in the cache and already cached predictions
+        cache_idx_dict = {
+            spec_idx: spec_idx * beam for spec_idx in range(batch)
+        }
+        # Keep already decoded peptides to avoid duplicates in cache
+        cache_seq_dict = {spec_idx: set() for spec_idx in range(batch)}
+        # Store peptide scores to replace lower score peptides in cache
+        # with higher score peptides during decoding.
+        cache_score_dict = {spec_idx: [[], []] for spec_idx in range(batch)}
+
+        return (
+            cache_scores,
+            cache_tokens,
+            cache_idx_dict,
+            cache_seq_dict,
+            cache_score_dict,
+        )
+
+    def _terminate_finished_beams(
+        self, tokens: torch.Tensor, precursors: torch.Tensor, idx: int
+    ) -> Tuple[List[int], torch.Tensor]:
+        """
+        Terminate beams exceeding precursor m/z tolerance and track all terminated beams
+
+        Parameters
+        ----------
+        tokens : torch.Tensor of shape (n_spectra* n_beams, max_length)
+            Output token of the model corresponding to amino acid sequences of peptides.
+        precursors : torch.Tensor of size (n_spectra* n_beams, 3)
+            The measured precursor mass (axis 0), precursor charge (axis 1), and
+            precursor m/z (axis 2) of each MS/MS spectrum.
+        idx : int
+            Index to be considered in the current decoding step
+
+        Returns
+        -------
+        finished_beams_idx : List[int]
+            Indices of all finished beams on tokens tensor
+        tokens : torch.Tensor of size (n_spectra* n_beams, max_length)
+            Output token of the model corresponding to amino acid sequences of peptides.
+
+        """
         # Terminate beams which exceed the precursor m/z
         for beam_i in range(len(tokens)):
             # Check only non-terminated beams
@@ -330,25 +422,67 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
         finished_beams_idx = torch.where(finished_beams_bool == True)[
             0
         ].tolist()
-        print(
-            "finished_beams_idx", len(finished_beams_idx), finished_beams_idx
-        )
-        return finished_beams_idx
+
+        return finished_beams_idx, tokens
 
     def _cache_finished_beams(
         self,
-        finished_beams_idx,
-        cache_idx_dict,
-        cache_seq_dict,
-        cache_score_dict,
-        cache_tokens,
-        cache_scores,
-        tokens,
-        precursors,
-        scores,
-        idx,
-    ):
+        finished_beams_idx: List[int],
+        cache_idx_dict: Dict[int, int],
+        cache_seq_dict: Dict[int, Set[str]],
+        cache_score_dict: Dict[int, List[List[Tuple[float, int]]]],
+        cache_tokens: torch.Tensor,
+        cache_scores: torch.Tensor,
+        tokens: torch.Tensor,
+        precursors: torch.Tensor,
+        scores: torch.Tensor,
+        idx: int,
+    ) -> Tuple[
+        Dict[int, int],
+        Dict[int, Set[str]],
+        Dict[int, List[List[Tuple[float, int]]]],
+    ]:
+        """
+        Cache terminated beams, group and order by fitting precursor m/z and confidence score
 
+        Parameters
+        ----------
+        finished_beams_idx : List[int]
+            Indices of all finished beams on tokens tensor
+        cache_idx_dict : Dict[int, int]
+            Next available tensor index to cache peptides for each spectrum.
+        cache_seq_dict : Dict[int,  Set[str]]
+            Set of decoded peptides for each spectrum.
+        cache_score_dict : Dict[int, List[List[Tuple[float, int]]]
+            Confidence score for each decoded peptide, separated as precursor m/z fitting vs not, for each spectrum.
+        cache_tokens : torch.Tensor of shape
+        (n_spectra* n_beams, max_length)
+            The token for each amino acid in cached peptides.
+        cache_scores : torch.Tensor of shape
+        (n_spectra* n_beams, max_length, n_amino_acids)
+            The score for each amino acid in cached peptides.
+        tokens : torch.Tensor of shape (n_spectra* n_beams, max_length)
+            Output token of the model corresponding to amino acid sequences of peptides.
+        precursors : torch.Tensor of size (n_spectra* n_beams, 3)
+            The measured precursor mass (axis 0), precursor charge (axis 1), and
+            precursor m/z (axis 2) of each MS/MS spectrum.
+         scores : torch.Tensor of shape (n_spectra*  n_beams, max_length, n_amino_acids)
+            Output scores of the model.
+        idx : int
+            Index to be considered in the current decoding step
+
+        Returns
+        -------
+        finished_beams_idx : List[int]
+            Indices of all finished beams on tokens tensor
+        cache_idx_dict : Dict[int, int]
+            Next available tensor index to cache peptides for each spectrum.
+        cache_seq_dict : Dict[int,  Set[str]]
+            Set of decoded peptides for each spectrum.
+        cache_score_dict : Dict[int, List[List[Tuple[float, int]]]
+            Confidence score for each decoded peptide, separated as precursor m/z fitting vs not, for each spectrum.
+
+        """
         beam = self.n_beams
         # Store finished beams in the cache
         if len(finished_beams_idx) > 0:
@@ -488,8 +622,40 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
         return cache_idx_dict, cache_seq_dict, cache_score_dict
 
     def _get_top_peptide(
-        self, cache_score_dict, cache_tokens, cache_scores, batch
-    ):
+        self,
+        cache_score_dict: Dict[int, int],
+        cache_tokens: torch.tensor,
+        cache_scores: torch.tensor,
+        batch: int,
+    ) -> Tuple[torch.tensor, torch.tensor]:
+        """
+        Return the peptide with highest confidence score, with fitting precursor m/z for each spectrum.
+        If no peptides within precursor m/z tolerance, return the highest scoring peptide among the non-fitting
+
+        Parameters
+        ----------
+        cache_score_dict : Dict[int, List[List[Tuple[float, int]]]
+            Confidence score for each decoded peptide, separated as precursor m/z fitting vs not, for each spectrum.
+        cache_tokens : torch.Tensor of shape
+        (n_spectra* n_beams, max_length)
+            The token for each amino acid in cached peptides.
+        cache_scores : torch.Tensor of shape
+        (n_spectra* n_beams, max_length, n_amino_acids)
+            The score for each amino acid in cached peptides.
+        batch: int
+            Number of spectra in the batch
+
+        Returns
+        -------
+        output_tokens : torch.Tensor of shape
+        (n_spectra, max_length)
+            The token for each amino acid in the output peptides.
+        output_scores : torch.Tensor of shape
+        (n_spectra, max_length, n_amino_acids)
+            The score for each amino acid in cached peptides.
+
+
+        """
         # Sizes:
         length = self.max_length + 1  # L
         vocab = self.decoder.vocab_size + 1  # V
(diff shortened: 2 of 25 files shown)


casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                    pred_mz = self.peptide_mass_calculator.mass(
+                        aa_list, charge=precursors[beam_i, 1].item()
+                    )
+                    delta_mass_ppm = (pred_mz - prec_mz) / prec_mz * 10**6
+                    if delta_mass_ppm > -self.precursor_mass_tol:
+                        tokens[beam_i][idx] = self.stop_token
+
+        # Get the list of indices for finished beams
+        finished_beams_bool = (tokens == self.stop_token).any(dim=1)
+        finished_beams_idx = torch.where(finished_beams_bool == True)[
```
bittremieux, line 422: **issue:** `== True` is not necessary.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+            0
+        ].tolist()
```
bittremieux, line 424: **question:** Can we keep this as a tensor and avoid the memory copies (slow)?

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+            Next available tensor index to cache peptides for each spectrum.
+        cache_seq_dict : Dict[int,  Set[str]]
+            Set of decoded peptides for each spectrum.
+        cache_score_dict : Dict[int, List[List[Tuple[float, int]]]
+            Confidence score for each decoded peptide, separated as precursor m/z fitting vs not, for each spectrum.
+
+        """
+        beam = self.n_beams
+        # Store finished beams in the cache
+        if len(finished_beams_idx) > 0:
```
bittremieux, line 488: **issue:** This `if` is not necessary; if the list is empty the iterator on the next line just won't run. Useful, because this method has a lot of indentations.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+            for i in finished_beams_idx:
+                spec_idx = i // beam  # find the starting index of the spectrum
+                insert_idx = cache_idx_dict[spec_idx]
+
+                # Only add beams with matching mass to the cache
+                prec_mz = precursors[i, 2].item()
+                aa_list = [
```
bittremieux, line 495: **question:** We're decoding the AAs and calculating the mass both when terminating and caching beams. Can we avoid doing this twice?

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                    self.decoder._idx2aa.get(idx_aa.item(), "")
+                    for idx_aa in tokens[i][:idx]
+                ]
+
+                # Check if dummy token predicted
+                if "" not in aa_list:
+                    # Remove stop token as needed, i.e. if mass filter induced termination
+                    if "$" in aa_list:
```
bittremieux, line 503: **suggestion:** Maybe the stop token and dummy token should be a member variable, rather than this "magic" string.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+
+                # Only add beams with matching mass to the cache
+                prec_mz = precursors[i, 2].item()
+                aa_list = [
+                    self.decoder._idx2aa.get(idx_aa.item(), "")
+                    for idx_aa in tokens[i][:idx]
+                ]
+
+                # Check if dummy token predicted
+                if "" not in aa_list:
```
bittremieux, line 501: **question:** In which situation is a predicted AA just empty?

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                    # Remove stop token as needed, i.e. if mass filter induced termination
+                    if "$" in aa_list:
+                        aa_list.remove("$")
+
+                    pep_str = "".join(aa_list)
+                    pred_mz = self.peptide_mass_calculator.mass(
+                        aa_list, charge=precursors[i, 1].item()
+                    )
+                    delta_mass_ppm = (pred_mz - prec_mz) / prec_mz * 10**6
+                    isPrecursorFit = (
```
bittremieux, line 511: **issue:** Use snake case as per the Python style conventions.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                        abs(delta_mass_ppm) <= self.precursor_mass_tol
```
bittremieux, line 512: **issue:** Compare with `<` for consistent behavior with `on_predict_epoch_end`. These are floats anyway, so `<` and `<=` are semantically identical.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                            cache_scores[insert_idx, :, :] = scores[i, :, :]
+                            cache_idx_dict[spec_idx] += 1  # move the pointer
+                            cache_seq_dict[spec_idx].add(pep_str)
+                            smx = self.softmax(scores)[i, :, :]
+
+                            aa_scores = [
+                                smx[idx][self.decoder._aa2idx[aa]].item()
+                                for idx, aa in enumerate(aa_list)
+                            ]
+                            pep_score = np.mean(aa_scores)
```
bittremieux, line 529: **suggestion:** It might be useful to extract this behavior to a AA-scores-to-peptide-scores function, so that if we want to change it, we can easily do it in a single location.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+
+                            # Cache peptides with fitting and non-fitting precursor mass separately
+                            if isPrecursorFit:
```
bittremieux, line 532: **suggestion:** Can be simplified as:
```
cache_i = not isPrecursorFit
heappush(cache_score_dict[spec_idx][cache_i], (pep_score, insert_idx))
```

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                                    (pep_score, insert_idx),
+                                )
+                            else:
+                                heappush(
+                                    cache_score_dict[spec_idx][1],
+                                    (pep_score, insert_idx),
+                                )
+
+                        else:
+                            smx = self.softmax(scores)[i, :, :]
```
bittremieux, line 544: **suggestion:** This behavior is duplicated in the `if` and the `else`, just do it once outside.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                            ]
+                            new_score = np.mean(aa_scores)
+
+                            # Cache fitting peptide either if non-fitting cached
+                            # or higher confidence than lowest scoring fitting peptide cached
+                            if isPrecursorFit:
+                                # Check if any non-fitting peptide cached
+                                if len(cache_score_dict[spec_idx][1]) > 0:
+                                    (
+                                        pop_pep_score,
```
bittremieux, line 557: **todo:** No need to capture the results from the `heappop` because they're not used further.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                                        cache_score_dict[spec_idx][0],
+                                        (new_score, pop_insert_idx),
+                                    )
+
+                                # Check if any fitting peptide cached
+                                elif len(cache_score_dict[spec_idx][0]) > 0:
+                                    (
+                                        pop_pep_score,
+                                        pop_insert_idx,
+                                    ) = heappop(cache_score_dict[spec_idx][0])
```
bittremieux, line 576: **suggestion:** Can we rewrite the succession of `heappush` and `heappop` more efficiently using `heappushpop`? Right now we're also popping and then pushing the same prediction back on the heap if the score was lower, which is a bit silly.

You can more nicely get the smallest element of a heap using `heap[0]`. That way you could "peek" at the value and only do a `heappushpop` if needed.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                                        heappush(
+                                            cache_score_dict[spec_idx][0],
+                                            (
+                                                pop_pep_score,
+                                                pop_insert_idx,
+                                            ),
+                                        )
+                            # Cache non-fitting peptide if higher confidence than lowest scoring non-fitting peptide cached
+                            else:
+                                if len(cache_score_dict[spec_idx][1]) > 0:
```
bittremieux, line 598: **suggestion:** Update the heap behavior similarly as above.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+            scores = einops.rearrange(scores, "B L V S -> (B S) L V")
+            tokens = einops.rearrange(tokens, "B L S -> (B S) L")
+
+            # Terminate beams exceeding precursor m/z tolerance and track all terminated beams
+            finished_beams_idx, tokens = self._terminate_finished_beams(
+                tokens, precursors, idx
+            )
+
+            # Cache terminated beams, group and order by fitting precursor m/z and confidence score
+            (
```
bittremieux, line 274: **suggestion:** `cache_idx_dict`, `cache_seq_dict`, and `cache_score_dict` are passed by reference in the function, so you don't necessarily need to return them. I'd rewrite this function without a return value to simplify the code a bit.

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+        output_scores = torch.zeros(batch, length, vocab)
+        output_scores = output_scores.type_as(cache_scores)
+        output_scores[output_scores == 0] = torch.nan
+        output_tokens = torch.zeros(batch, length)
+        output_tokens = output_tokens.type_as(cache_tokens)
+
+        # Return the top scoring peptide (fitting precursor mass if possible)
+        for spec_idx in range(batch):
+            cached_fitting, cached_nonfitting = cache_score_dict[spec_idx]
+            if len(cached_fitting) != 0:
```
bittremieux, line 673: **suggestion:** Can be slightly simplified as:
```
cache = cached_fitting if len(cached_fitting) > 0 else cached_nonfitting
_, top_score_idx = max(cache, key=lambda item: item[0])
```

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                                smx[idx][self.decoder._aa2idx[aa]].item()
+                                for idx, aa in enumerate(aa_list)
+                            ]
+                            new_score = np.mean(aa_scores)
+
+                            # Cache fitting peptide either if non-fitting cached
+                            # or higher confidence than lowest scoring fitting peptide cached
+                            if isPrecursorFit:
+                                # Check if any non-fitting peptide cached
+                                if len(cache_score_dict[spec_idx][1]) > 0:
```
bittremieux, line 555: **todo:** Can we add a few unit tests that cover the below situations without test coverage?

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+        ).any(axis=1)
+        # Mask out the index '0', i.e. padding token, by default
+        finished_mask[:, :beam] = True
+
+        # Figure out the top K decodings
+        _, top_idx = torch.topk(
+            step_scores.nanmean(dim=1) * (~finished_mask).int().float(),
+            beam,
+        )
+        V_idx, S_idx = np.unravel_index(top_idx.cpu(), (vocab, beam))
```
bittremieux, line 762: **Issue:** Don't start variable names with a capital letter, as per the Python style convention.

casanovo/denovo/model.py
```diff
@@ -411,11 +1008,14 @@ def on_predict_epoch_end(
...
-                    # Take the scores of the most probable amino acids.
-                    top_aa_scores = torch.max(
-                        aa_scores[1 : len(peptide_tokens) + 1], axis=1
-                    )[0]
-                    peptide_score = torch.mean(top_aa_scores).detach().item()
+                    # Take the scores corresponding to the predicted amino acids.
+                    top_aa_scores = [
+                        aa_scores[idx][self.decoder._aa2idx[aa]].item()
+                        for idx, aa in enumerate(
+                            list(reversed(peptide_tokens))
```
bittremieux, line 1015: **issue:** No need for the `list` conversion.

tests/test_unit.py
```diff
@@ -99,8 +100,163 @@ def request(self, *args, **kwargs):
...
 
     model = Spec2Pep()
     assert model.tb_summarywriter is None
+
+
+def test_beam_search_decode():
+    """
+    Test that beam search decoding and its sub-functions as intended during inference
+    """
+    model = Spec2Pep(n_beams=4)
```
bittremieux, line 117: **issue:** Can we have some small tests with different numbers of beams (1 should be there for sure).

bittremieux (changes requested):
Nice work. This is non-trivial functionality to implement.

Here is some initial feedback. I'll probably want to do another thorough review after these comments have been addressed.

Some comments are relevant to multiple places in the code, but I haven't repeated the same thing multiple times, so be a bit mindful of that.

casanovo/config.yaml
```diff
@@ -65,6 +65,7 @@ weight_decay: 1e-5
 # Training/inference options.
 train_batch_size: 32
 predict_batch_size: 1024
+n_beams: 1
```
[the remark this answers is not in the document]
melihyilmaz: `n_beams` determines both the number of beams explored and total number of beams cached. Model currently only reports a single PSM regardless of this parameter. We can add functionality to report multiple PSMs later.

I implemented beams explored = beams cached both because it's straight-forward but also because that's the minimum number of peptide predictions we are guaranteed to come up with for a given number of beams, i.e. larger cache than number of beamse doesn't guarantee more PSMs (if/when we start reporting multiple PSMs)

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+                cache_tokens,
+                cache_scores,
+                tokens,
+                precursors,
+                scores,
+                idx,
+            )
+
+            # Stop decoding when all current beams are terminated
+            decoded = (tokens == self.stop_token).any(axis=1)
```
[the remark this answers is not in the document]
melihyilmaz: Decoding stops when _all_ beams are terminated.

casanovo/config.yaml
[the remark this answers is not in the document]
bittremieux: Beams explored = beams cached is fine I think, but we will want an option (as a future feature addition is fine) to report multiple PSMs per spectrum.

When considering only a single beam, the result is the same as for greedy decoding. So we probably want to increase the default number of beams a bit to actually get advantage of this functionality.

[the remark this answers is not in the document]
melihyilmaz: Raising it to 5 and opening an issue for multi-PSM output for future reference

casanovo/denovo/model.py
melihyilmaz: Will cache delta m/z to avoid duplication but couldn't come up with a straight forward way of caching the peptide to avoid idx2AA decoding twice i.e. beams can terminate when stop token predicted before precursor m/z is exceeded which causes all sorts of edge cases

casanovo/denovo/model.py
```diff
@@ -176,12 +193,592 @@ def forward(
...
+            # Check only non-terminated beams
+            if self.stop_token not in tokens[beam_i]:
+                # Finish if dummy predicted at the previous step
+                if tokens[beam_i][idx - 1] == 0:
+                    tokens[beam_i][idx - 1] = self.stop_token
+                # Check delta m/z and terminate the beam if exceeds the threshold
+                else:
+                    prec_mz = precursors[beam_i, 2].item()
+                    aa_list = [
+                        self.decoder._idx2aa.get(idx_aa.item(), "")
```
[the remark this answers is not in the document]
melihyilmaz: Switched to `detokenize`

casanovo/denovo/model.py
melihyilmaz: It was a remnant from earlier debugging, now removed.

casanovo/denovo/model.py
melihyilmaz: Extracted to `_aa_to_pep_score()`

casanovo/denovo/model.py
melihyilmaz: Added more test cases to cover replacement of 1) non-fitting with fitting and 2) non-fitting with non-fitting  in the cache

tests/test_unit.py
melihyilmaz: Added `n_beams=1` but couldn't think of other meaningful **small** tests to cover other beam sizes.

casanovo/denovo/model.py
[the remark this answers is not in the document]
melihyilmaz: Regex logic in `peptide_mass_calculator.mass` can't handle cases where multiple N-terminal modifications are predicted (e.g. `-17.027+42.011`) if the peptide is `str` so it's useful to decode this way to have an AA list. Made this change elsewhere in `model.py` as well.

casanovo/denovo/model.py
```diff
@@ -195,7 +196,18 @@ def forward(
             spectra.to(self.encoder.device),
             precursors.to(self.decoder.device),
         )
-        return [self.decoder.detokenize(t) for t in tokens], aa_scores
+        peptides = []
```
bittremieux, line 199: **issue:** I don't understand this change. This copies the `detokenize` code from depthcharge and does exactly the same as what the previous one-line statement did, except that it's a list rather than a string.

1. Why is this change necessary? Both list and strings are iterables.
2. Even if this change is necessary, duplicating 95% of functionality from depthcharge is not the best solution.

casanovo/denovo/model.py
```diff
@@ -418,8 +433,10 @@ def _terminate_finished_beams(
                 else:
                     precursor_mz = precursors[beam_i, 2].item()
                     precursor_charge = precursors[beam_i, 1].item()
-                    peptide_seq = self.decoder.detokenize(tokens[beam_i][:idx])
-
+                    peptide_seq = [
```
bittremieux, line 436: **issue:** Similar question as above.

casanovo/denovo/model.py
```diff
@@ -444,6 +461,37 @@ def _terminate_finished_beams(
                             abs(d) < self.precursor_mass_tol
                             for d in delta_mass_ppm
                         )
+
+                        # Don't terminate if m/z difference is by NH3 loss
```
bittremieux, line 465: **issue:** This solution seems sub-optimal because it considers NH3 loss as a special token, whereas there are different N-terminal modifications as well. Why did NH3 loss get special treatment? I adds a decent amount of extra complexity and I don't understand why this is needed (not documented).

casanovo/denovo/model.py
```diff
@@ -444,6 +461,37 @@ def _terminate_finished_beams(
...
+                        if (
+                            exceeds_precursor_mz_tol
+                            and not is_within_precursor_mz_tol
```
bittremieux, line 468: **nitpick:** This is always the case if `exceeds_precursor_mz_tol` is True, hence not really needed.

casanovo/denovo/model.py
```diff
@@ -444,6 +461,37 @@ def _terminate_finished_beams(
...
+                            and is_nh3_loss_in_vocab
+                        ):
+                            alt_peptide_seq = peptide_seq + ["-17.027"]
```
bittremieux, line 471: **issue:** NH3 loss is an N-terminal modification, whereas here it's added at the end (i.e. C-terminal). So either the modification appears in the wrong location in the peptide string, or the peptide strings hasn't been reversed yet?
For mass calculation I guess the position doesn't really matter, but at the very least it's confusing. Worst case scenario it results in hard-to-diagnose bugs now or in the future.

casanovo/denovo/model.py
```diff
@@ -444,6 +461,37 @@ def _terminate_finished_beams(
...
+                                for isotope in range(
+                                    self.isotope_error_range[0],
+                                    self.isotope_error_range[1] + 1,
+                                )
+                            ]
+                            is_precursor_mz_tol_w_loss = any(
+                                abs(d) < self.precursor_mass_tol
+                                for d in delta_mass_ppm
+                            )
+                            exceeds_precursor_mz_tol = (
```
bittremieux, line 491: **nitpick:** Combine the previous statement with this one, i.e. `not any(...)`. Just doing `not` of another boolean is a bit silly.

bittremieux (changes requested):
It's looking pretty good. I have a few comments / requests for clarification. I especially don't understand the latest NH3 loss-related changes.

casanovo/denovo/model.py
```diff
@@ -171,16 +178,613 @@ def forward(
...
+        scores[:, :1, :, :] = einops.repeat(pred, "B L V -> B L V S", S=beam)
+
+        # Make precursors and memories the right shape for decoding:
+        precursors = einops.repeat(precursors, "B L -> (B S) L", S=beam)
+        mem_masks = einops.repeat(mem_masks, "B L -> (B S) L", S=beam)
+        memories = einops.repeat(memories, "B L V -> (B S) L V", S=beam)
+
+        # The main decoding loop:
+        for idx in range(1, self.max_length + 1):
+            scores = einops.rearrange(scores, "B L V S -> (B S) L V")
```
bittremieux, line 272: **question:** Can this be moved outside of the loop?

casanovo/denovo/model.py
melihyilmaz: NH3 loss is the only token with negative mass. In the previous implementation, early stopping kicked in the even if the beam was expecting a NH3 loss as the last token because m/z tolerance would be exceeded. I couldn't think of a more elegant way of covering this edge case and avoiding m/z threshold based stopping if the m/z margin is equivalent to NH3 loss

casanovo/denovo/model.py
```diff
@@ -171,16 +178,613 @@ def forward(
...
+                tokens[~decoded, :idx],
+                precursors[~decoded, :],
+                memories[~decoded, :, :],
+                mem_masks[~decoded, :],
+            )
+
+            # Find top-k beams with highest scores and continue decoding those
+            scores, tokens = self._get_topk_beams(scores, tokens, batch, idx)
+
+        return self.softmax(output_scores), output_tokens
```
bittremieux, line 317: **issue:** In the unrealistic situation that decoding of one or more beams didn't finish within `max_length` steps, `output_scores` and `output_tokens` will be undefined. Should we move the `return` up to within the `if decoded.all()` and appropriately handle the edge case when decoding didn't fully finish here?

casanovo/denovo/model.py
```diff
@@ -171,16 +178,613 @@ def forward(
...
+        Returns
+        -------
+        finished_beams_idx : torch.Tensor
+            Indices of all finished beams on tokens tensor.
+        tokens : torch.Tensor of size (n_spectra * n_beams, max_length)
+            Output token of the model corresponding to amino acid sequences.
+        """
+        # Check if N-terminal NH3 loss is in the vocabulary
+        is_nh3_loss_in_vocab = (
+            self.peptide_mass_calculator.masses.get("-17.027", "") != ""
```
bittremieux, line 412: **todo:** This can be simplified by `"-17.027" in self.peptide_mass_calculator.masses`.

casanovo/denovo/model.py
```diff
@@ -171,16 +178,613 @@ def forward(
...
+        finished_mask = (
+            einops.rearrange(extended_prev_tokens, "B L V S -> B L (V S)")
+            == self.stop_token
+        ).any(axis=1)
+        # Mask out the index '0', i.e. padding token, by default
+        finished_mask[:, :beam] = True
+
+        # Figure out the top K decodings
+        _, top_idx = torch.topk(
+            step_scores.nanmean(dim=1) * (~finished_mask).int().float(), beam
```
bittremieux, line 772: **question:** Weird to convert to int and then immediately to float.

casanovo/denovo/model.py
bittremieux: Ok, I didn't think/realize that NH3 loss is the only negative mass. Then it is indeed a bit of a special case, but we can try to generalize the code a little bit after the remaining comments have been addressed.

casanovo/denovo/model.py
melihyilmaz: Regex logic in `peptide_mass_calculator.mass` downstream can't handle cases where multiple N-terminal modifications are predicted back to back (e.g. -17.027+42.011) if the peptide is `str` rather than list so it's useful to decode this way to have an AA list. We can implement a new regex to convert `str` to list here or `peptide_mass_calculator.mass` can be modified in DepthCharge .

casanovo/denovo/model.py
```diff
@@ -176,12 +191,602 @@ def forward(
...
+                                precursor_mz,
+                                precursor_charge,
+                                isotope,
+                            )
+                            for isotope in range(
+                                self.isotope_error_range[0],
+                                self.isotope_error_range[1] + 1,
+                            )
+                        ]
+                        exceeds_precursor_mz_tol = any(
```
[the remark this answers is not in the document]
melihyilmaz: I think this makes sense but should the same logic apply to `is_within_precursor_mz_tol` as well? It wouldn't have downstream implications for the beam as long as`exceeds_precursor_mz_tol = False` but it'd still lead to discrepancy between the two.

casanovo/denovo/model.py
melihyilmaz: They can both be True at the same time (`exceeds_precursor_mz_tol` is True for any mass delta higher than lower bound - so it includes the tolerance range), in which case we wouldn't want to go check for NH3 loss.

casanovo/denovo/model.py
melihyilmaz: How should we handle case if decoding didn't fully finish?

casanovo/denovo/model.py
melihyilmaz: Moved it out of the loop and added reshaping inside `_get_topk_beams `

casanovo/denovo/model.py
melihyilmaz: Peptide wasn't reversed at that point. Changed to this have peptide reversed before hand and modification added to N-terminus. Can revisit after `detokenize` resolved

casanovo/denovo/model.py
bittremieux: I prefer to keep the (de)tokenizing functionality fully on the depthcharge side, rather than mixing it in the different codebases. I suggest to revert the changes here and directly open a PR in depthcharge that will need to be integrated prior to merging this PR.

casanovo/denovo/model.py
bittremieux: My suggestion: if at least one beam finished, select from the finished beams. If no beam finished at all, just return dummy values (zero). We can write a unit test for this behavior by setting `max_length` to some ridiculously small number to verify that our implementation is correct, although this will likely never occur in reality (or at least it shouldn't).

melihyilmaz: Made the changes you requested and implemented a test controlling if a finished beam is return if there are any. However, couldn't find a straightforward way to test if dummy values returned when `max_length` very small. Should I initialize a fixed model with known output? If not feel free to resolve the conversation

casanovo/denovo/model.py
melihyilmaz: Reverted to `detokenize` and updated depthcharge dependency

melihyilmaz:
Added a small fix to negative mass-aware termination (to ensure we're not terminating a beam if doesn't exceed tolerance under _any_ negative mass token) and unit tests covering that functionality. Only that and 1 other unresolved conversation above remains.

casanovo/denovo/model.py
melihyilmaz: @bittremieux in case you haven't seen my earlier comment

casanovo/denovo/model.py
```diff
@@ -408,21 +875,27 @@ def on_predict_epoch_end(
...
-                        aa_scores[1 : len(peptide_tokens) + 1], axis=1
-                    )[0]
-                    peptide_score = torch.mean(top_aa_scores).detach().item()
+                for spectrum_i, precursor, aa_tokens, aa_scores in zip(*step):
+                    # Omit stop token.
+                    aa_tokens = aa_tokens[1:]
+                    peptide = "".join(aa_tokens)
+                    # If this is a non-finished beam (after exceeding
+                    # `max_length`), return a dummy (empty) peptide.
+                    if len(peptide) == 0:
```
bittremieux, line 884: **question:** I don't understand these two lines. `peptide` is created from `aa_tokens` and only has length 0 if `aa_tokens` is empty, so this sets it from an empty list to an empty string? Both are empty iterables, so this code doesn't really do anything?

melihyilmaz: If `aa_tokens` has a number of empty strings `len(aa_tokens)>0` which we try to iterate over below. The list of empty strings arise from dummy tokens.

bittremieux: Got it, thanks for the explanation. Maybe we should add a comment with that info for future reference.

However, I think that the code will crash in some of the following lines in that case. For example, if `aa_tokens` is empty, `top_aa_scores` will also be an empty list, and taking the mean when calculating the peptide score will crash.

So I don't think we have a test for this situation yet? And maybe rather than setting `aa_scores` to empty, we should not do any of the following code and directly give a 0-score to the mzTab writer?

bittremieux: Can you also verify that the AA scores and peptide sequence are correctly ordered? Right now we take the peptide from the `aa_tokens` directly in line 881, but we ultimately reverse the AA scores before writing them in line 925. It seems to me that those two operations are not in sync with each other.

melihyilmaz: re: empty list, as `np.mean` returns NaN without crashing the method doesn't crash and outputs empty string for the peptide and NaN for score in the mzTab. I think this can stay as is but I can implement a form of zero output if you think that's better.

melihyilmaz: re: AA scores and tokens order, I corrected the bug introduced in [053967e](https://github.com/Noble-Lab/casanovo/pull/87/commits/053967edc80b31c998dcb446c7aae861d04b3fd8), refactored the code and added tests for this functionality.

casanovo/denovo/model.py
melihyilmaz: I added a test case.

casanovo/denovo/model.py
bittremieux: Woopsie. Thanks for catching that!

I was also too fast in checking `np.mean`, which indeed doesn't crash on empty input but gives a `RuntimeWarning`. I saw a stack trace and didn't look closely enough. But this works then.

casanovo/denovo/model.py
```diff
@@ -460,6 +928,52 @@ def on_predict_epoch_end(
...
+        aa_tokens : List[str]
+            Amino acid tokens of the peptide sequence.
+        peptide_score : str
+            Peptide-level confidence score.
+        aa_scores : str
+            Amino acid-level confidence scores for the predicted sequence.
+        """
+        # Omit stop token.
+        aa_tokens = aa_tokens[1:] if self.decoder.reverse else aa_tokens[:-1]
+        peptide = "".join(aa_tokens)
```
bittremieux, line 957: **question:** One final confusion about the order of the tokens and AAs. Here we create `peptide` from `aa_tokens` without reversing them, whereas in line 969 the tokens are reversed prior to getting the AA scores. Doesn't that contradict itself?
Although I guess the scores are reversed twice in lines 969-970, but why do `aa_scores` only need the be reversed once?

melihyilmaz: `aa_tokens` arrives in the correct order in any case (except for the position of stop token) but aa scores arrive reversed `if self.decoder.reverse = True` so that's why we a) reverse `aa_tokens` to correspond to the order of scores for retrieval and then b) retrieved aa scores are reversed to correspond to the N- to C-terminus order. Let me know if I'm missing anything.

bittremieux: Ok, rather than reversing `aa_tokens` twice, wouldn't it be easier to reverse `aa_scores` only once? E.g.:
```
top_aa_scores = [
    aa_score[self.decoder._aa2idx[aa_token]].item()
    for aa_score, aa_token in zip(aa_scores[::step], aa_tokens)
]
```

Slightly less code, but also less confusing to understand what's happening I think.

melihyilmaz: `aa_scores` is a tensor with size `(max_length, vocab)` so both the snippet you shared doesn't work as is and also reversal of the non-empty positions in the tensor wasn't as straightforward for me.

bittremieux: That makes total sense! Thanks.

bittremieux (approved):

PR #94, opened 2022-11-08
Fix weights version

bittremieux:
Small bugfix to not use model weights with an incorrect major version number.


wfondrie (approved):
These changes look good to me!

PR #97, opened 2022-11-11
Add unit test for evaluation metrics

melihyilmaz:
Added a unit test for for peptide and amino acid-level evaluation metrics after reviewing `evaluate.py`.


wfondrie (approved):
Looks good to me!

PR #104, opened 2022-12-01
Update README.md

wsnoble:
As discussed on Slack, I shortened the README file to remove redundant content with the readthedocs.


wfondrie (commented):
Are you sure you don't want to leave at least a little information about what Casanovo is in the README? For example, I have a short description and abbreviated getting started for [mokapot](https://github.com/wfondrie/mokapot/blob/master/README.md).

wsnoble:
I don't actually want to include the abbreviated "getting started" section, because I want to make sure people actually click through to the documentation before trying to install it. I'm open to putting a bit more about what it is, but I feel like it's already pretty well covered by "De Novo Mass Spectrometry Peptide Sequencing with a Transformer Model."

wfondrie (approved):

PR #105, opened 2022-12-02
Change PSM ID with identifier in peak file

### LINKED ISSUE #70: Track input spectra origin
Currently predicted PSMs are identified by the index of their spectra in the data loader. However, when predicting from multiple input files, this doesn't include information from which file the spectra came, pretty much making the index useless.

Can we track the origin of the input spectra, i.e. using filename and scan number? This probably needs to be modified in the index in depthcharge, and should then be passed through when predicting (but no need to move that information into a tensor on the GPU) so that it is available when writing the output results.

melihyilmaz:
Fixes #70.

Addressing [issue](https://github.com/Noble-Lab/casanovo/issues/70), we change PSM ID in the output mztab files, which only denotes the order in output file, to PSI standard identifier used in the peak file.


tests/test_integration.py
```diff
@@ -20,3 +22,12 @@ def test_denovo(mgf_small, tmp_path, monkeypatch):
         standalone_mode=False,
     )
     assert output_filename.is_file()
+
+    with open(output_filename) as f_in:
+        for skiprows, line in enumerate(f_in):
+            if line.startswith("PSH"):
+                break
```
wfondrie, line 29: **nitpick:** Not required, but I would add a newline here.

wfondrie (commented):
These changes look good to me - we just need to figure out why tests failed on GitHub.

wfondrie:
The first line in the Pytest error seems to indicate that there are no PSMs written in the output file:
```
self = Series([], Name: PSM_ID, dtype: object), key = 0

    def __getitem__(self, key):
        check_deprecated_indexers(key)
        key = com.apply_if_callable(key, self)
    
        if key is Ellipsis:
            return self
    
        key_is_scalar = is_scalar(key)
        if isinstance(key, (list, tuple)):
            key = unpack_1tuple(key)
    
        if is_integer(key) and self.index._should_fallback_to_positional:
>           return self._values[key]
E           IndexError: index 0 is out of bounds for axis 0 with size 0

/opt/hostedtoolcache/Python/3.10.8/x64/lib/python3.10/site-packages/pandas/core/series.py:978: IndexError
```

casanovo/data/datasets.py
```diff
@@ -84,7 +84,31 @@ def __getitem__(self, idx) -> Tuple[torch.Tensor, float, int, str]:
...
+            The index of the MS/MS spectrum within the SpectrumIndex.
+
+        Returns
+        -------
+        ms_data_file : str
+            The peak file from which the MS/MS spectrum was originally parsed.
+        identifier : str
+            The MS/MS spectrum identifier, per PSI recommendations.
+        """
+        with self.index:
```
bittremieux, line 110: **question:** Could it be suboptimal that we always need to open a context manager to get a spectrum identifier? I don't really have a better solution though. What do you think @wfondrie?

bittremieux (approved):
I added a test to verify that spectrum identifiers from mzML files are correct as well. Unlike MGF files, mzML reports scan numbers rather than indexes. I don't know a library that can simulate mzXML files, so we don't explicitly test that behavior. But mzXML is quite similar to mzML, so I think our tests are sufficient.

The added behavior works as intended, and unit and integration tests are all successful locally. The failing test is because the generated mzTab indeed doesn't contain PSMs. This is not related to changes in this PR, but rather because of issues with multiprocessing (again). Running with a GPU works, but CPU-only (as the tests are run using GitHub Actions) seems problematic. This is not a new issue, our previous tests just didn't capture it.

Changing the [strategy](https://github.com/Noble-Lab/casanovo/blob/main/casanovo/denovo/model_runner.py#L348) to `None` and the [number of devices](https://github.com/Noble-Lab/casanovo/blob/main/casanovo/denovo/model_runner.py#L367) to `1` actually makes the unit tests run correctly here. But that removes all multiprocessing as well of course. I suggest that we merge this PR, rather than keep it dangling for an indefinite amount of time, and address the CPU-only multiprocessing in a follow-up PR.

What do you think @wfondrie?

bittremieux:
Although my approval might have been a bit hasty. We still need to add an extra test: predicting from multiple files. In that case, the different files should be reported as `ms_run[x]-location` in the mzTab header, and the PSMs should refer to the correct ms_run index. in the `spectra_ref` column This behavior is not correct yet in the current code.

wfondrie:
I agree with @bittremieux!

bittremieux:
Ok, I fixed the mapping from PSMs to input files. Can you do a new review before merging @wfondrie?

melihyilmaz:
I reviewed Wout's changes and tests pass locally on a GPU machine - good to merge!

PR #113, opened 2023-01-09
Preprint info

bittremieux:
Add link to the new preprint.


wfondrie (approved):
Looks good to me. My only suggestion is change the preceding "publication" to "publications"

Issue #122, opened 2023-01-17
Inconsistent metrics result between eval and denovo mode, and possible reason


irleader:
This is a follow-up of issue  #111, thanks to all the information provided, I am able to find a possible bug and its reason.

I did experiment with weights "pretrained_excl_yeast.ckpt", deepnovo original raw yeast dataset, and Casanovo version 3.1.0.

In eval mode (results displayed on console):
**aa_precision: 0.7057459339992785
aa_recall: 0.7071716086435299**
pep_recall: 0.5046985050978404

In denovo mode (results manually calculated from output mztab file):
**aa_precision: 0.6827751198617065
aa_recall: 0.6843725335842246**
pep_recall: 0.5046985051027288

There is difference between results from eval and denovo mode (aa_precision, aa_recall), so I studies model.py and its pytorch lightning implementation. 

I first excluded the possibility that this may be caused by predictions without '$' or with multiple '$$' (it turns out there are no such predictions), because [this line of code](https://github.com/Noble-Lab/casanovo/blob/ed2490a635b39236ae0172c95182e95604394bf6/casanovo/denovo/model.py#L956) removes the first position of prediction without checking whether it is '$', I am referring to v3.2.0, the code in v3.1.0 is similar.

Then I checked validation part of the code, and find the inconsistence might be caused by the way [pytorch lightning accumulates metrics from each step and calculates per-epoch metrics](https://github.com/Noble-Lab/casanovo/blob/ed2490a635b39236ae0172c95182e95604394bf6/casanovo/denovo/model.py#L807-L810). So I manually accumulates the prediction data from each step and calculate metrics for one epoch, and find it is the same as the results from denovo mode, which verifies my guess. For manual calculation, I use methods from evaluate.py.

Manual calculation from validation_step() output:
**aa_precision: 0.6827751198617065
aa_recall: 0.6843725335842246**
pep_recall: 0.5046985051027288

 In order to further verify, I tried with deepnovo raw ricebean dataset, and see the same pattern.

In eval mode (results displayed on console):
   **aa_precision:     0.5835929214378351
     aa_recall:      0.583695772587527**
    pep_recall:      0.4223428193208169

In denovo mode (results manually calculated from output mztab file):
   **aa_precision:     0.576437557764845  
     aa_recall:      0.5758105380102586**   
    pep_recall:      0.4223428193248386

Manual calculation from validation_step() output:
   **aa_precision:     0.576437557764845
[...]

PR #123, opened 2023-01-17
Check for stop token

bittremieux:
Because beams can be terminated when the precursor mass is exceeded, peptide sequences might not necessarily end with a stop token.

Reported in #122.


melihyilmaz (approved):
We're adding a stop token to all terminated beams (i.e. both when stop token is predicted and    with precursor m/z filter) but still could be useful to have a simple check for potential future use cases.

Issue #112, opened 2023-01-08
Change printed output to reflect terminology in the paper


melihyilmaz:
https://github.com/Noble-Lab/casanovo/blob/f0f2505cbadd923d3048547215b2c0356d41fbdf/casanovo/denovo/model.py#L810

PR #126, opened 2023-01-18
Align printed metrics with published terminology

melihyilmaz:
Minor PR addressing #112 which updates the printed evaluation metrics in the `eval` mode to agree with terms such as precision and coverage we use in the preprint and ICML paper.


bittremieux (changes requested):
Looks good, although the same changes should also be applied to functions and docstrings in `evaluate.py`.

melihyilmaz:
Good point! Fixed those now

bittremieux (approved):

PR #130, opened 2023-01-30
First Casanovo DB Search Iteration

VarunAnanth2003:
Output format is a `.csv` of lists of floats that correspond with the score between a spectra-peptide pair in the multi-annotated `.mgf`. It is always saved to `test.csv`.

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -1110,7 +1110,7 @@ def calc_match_score(
         for scores, true_index in zip(all_aa_pred, truth_indicies):
             aa_scores.append(scores[true_index].item())
         normalized_score = np.mean(
-            aa_scores
+            aa_scores  #! Convert to product and normalization by length, or alternatively use LogSoftmax
         )  # Normalize score along peptide length
         all_scores.append(normalized_score)
         per_aa_scores.append(aa_scores)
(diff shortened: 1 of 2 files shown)


casanovo/denovo/model.py
```diff
@@ -1059,14 +1059,16 @@ def predict_step(self, batch, *args):
             # Use _forward_step
```
justin-a-sanders, line 1059: I would factorize the batch construction here into a separate method and add a quick unit test for it - looks correct now, but we want to make sure it keeps working if/when we change file formats. Weird predictions stemming from peptides being matched to the wrong spectra/precursors would be pretty annoying to track down.

casanovo/denovo/model.py
```diff
@@ -1096,24 +1098,23 @@ def calc_match_score(
...
 
     for all_aa_pred, truth_indicies in zip(
         batch_all_aa_scores, truth_aa_indicies
     ):
         assert len(all_aa_pred) == len(
             truth_indicies
         )  # Ensure that length of score list and indexes to pull from are the same length
-        aa_score_sum = 0
+        aa_scores = []
         for scores, true_index in zip(all_aa_pred, truth_indicies):
```
justin-a-sanders, line 1110: Nit: Would probably be cleaner to get per_aa_scores all at once using a pytorch function call to get all the needed indexes (eg. using torch.gather) rather than the nested for loop. Could bring this method down to a couple of lines.

justin-a-sanders (approved):

VarunAnanth2003:
Made the requested fixes @justin-a-sanders, didn't do the NIT but will get to that later. Can we merge this?

justin-a-sanders (approved):

PR #132, opened 2023-02-02
Add training data information

### LINKED ISSUE #131: Availability of MassIVE-KB dataset
From the preprint, "MassIVE-KB contains a designated subset of 30,506,973 “high quality” PSMs, identified by applying a strict (∼0%) PSM-level FDR filter and then selecting at most 100 PSMs for each combination of peptide sequence and charge."

Did you download the original MassIVE knowledge base from https://massive.ucsd.edu/ProteoSAFe/static/massive-kb-libraries.jsp? Which specific library did you download from, is it Human HCD Spectral Library of v1?

Is it possible to share your processing pipeline (simply using PEAKS?) or your processed MassIVE-KB dataset? I would like to train Casanovo by myself. Thanks.

bittremieux:
Fixes #131.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -19,12 +19,6 @@ casanovo --help
 Additionally, you can use a configuration file to fully customize Casanovo.
 You can find the `config.yaml` configuration file that is used by default [here](https://github.com/Noble-Lab/casanovo/blob/main/casanovo/config.yaml).
 
-**I get a "CUDA out of memory" error when trying to run Casanovo. Help!**
-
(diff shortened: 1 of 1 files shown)


wsnoble (commented):
I think this looks good.

Longer term, I think we should resist the temptation to have our documentation consist of an unstructured list of FAQs. It would be nice to have topic headings, such as "Data sets" for this entry. But the current approach is OK for now.

Side question: can we put the 30M dataset on zenodo, or is it too large?

wsnoble (approved):

PR #134, opened 2023-02-09
Replace list comprehension with slicing

seantur:
Thanks for the great work on Casanovo!

I was doing some profiling and noticed that this list comprehension was taking quite a bit of time, at least for my workloads. Replacing it with slicing offers a pretty significant speed up. I'm not totally sure if the dtype needs to be `float64` - I just made the type match the implementation of the v3.2.0 release.

I'm happy to make any changes.

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -520,8 +520,13 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
             if is_peptide_cached:
                 continue
             smx = self.softmax(scores)
-            aa_scores = np.asarray(
-                [smx[i, j, k].item() for j, k in enumerate(pred_seq)]
+
+            aa_scores = (
+                smx[i, np.arange(len(pred_seq)), pred_seq]
+                .to("cpu")
+                .detach()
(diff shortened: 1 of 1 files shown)


bittremieux (approved):
Hi Sean, this looks great. Thank you for the contribution.

Moving the full array from the GPU to CPU in a single operation, rather than individual elements during the list comprehension, is indeed considerably more efficient.

bittremieux (changes requested):
There does still seem to be an issue with the unit tests though. Do you mind checking what's going on there and making sure that everything works?

seantur:
Hi Wout,

Thanks for the feedback. I was in fact missing a `len`. The unit tests also pointed out that if the tensor requires grad, that you need to `detach` before calling numpy.

The unit tests are now passing locally for me.

Happy to make any more changes if needed.

bittremieux (commented):
Looks good. I'm going to hold off on merging the PR for the moment though because I've been doing some refactoring of the prediction caching functionality, which will cause some merge conflicts with these changes. I might directly incorporate this in my PR instead.

PR #136, opened 2023-02-13
Don't predict short peptides

### LINKED ISSUE #107: Casanovo confidently predicts short peptide sequences (< 5 aa)
Casanovo will occasionally very confidently predict very short peptides, with high mass differences. 
Attached is a zip archive with an .mgf with 72 spectra that the non-enzymatic model predicts to be less than 5 aa long, as well as the predictions from that model. Base model predictions are included as well, though the base model fares much better with these spectra.

[short-predictions.zip](https://github.com/Noble-Lab/casanovo/files/10173443/short-predictions.zip)

### LINKED ISSUE #115: Refuse to predict peptides that are too short
We should add logic to the beam search so that stop tokens that yield predicts that are too short (i.e., the mass is less than the precursor minus the window size) are disallowed.

This is related to issue #107, and should be addressed after that bug is fixed.

bittremieux:
Rather than simply modifying the peptide score to account for low-scoring stop tokens after very confident prefixes in low-ranking beams, I ended up refactoring the beam search and peptide caching functionality quite a bit.

- The score of the stop token is now included when calculating the peptide score. If no stop token is predicted, an additional "amino acid" with score `0` is added for the purpose of calculating the peptide score. (The score of the stop token is not reported in the mzTab output.)
- Rather than calculating the peptide score when evaluating the beam and when collecting the output, I removed this functionality duplication and compute the peptide score only once when evaluating the beam and then store it.
- This also moves the precursor m/z filter directly in the beam evaluation. Consequently, peptide caching could be simplified significantly. There's now only a single cache for each beam (as opposed to precursor fitting vs not-precursor fitting), because predictions that don't fit the precursor m/z immediately get penalized, and the predicted peptides are directly returned from `beam_search_decode` rather than the raw tokens.
- I added a parameter that controls the minimum length of the predicted peptides (default: 7). If beams include the stop token before this minimum peptide length is reached, they get discarded.

The upshot is that very short peptides are no longer predicted and predictions that don't include a stop token (i.e. because they exceed the precursor m/z before the stop token occurs) are penalized.

@melihyilmaz:
- I would appreciate a thorough review of the code changes.
- The unit tests currently don't fail, but some of them are commented out or don't test anything useful. Can you have a look and updated the commented-out tests and add relevant new tests to verify that the new functionality works as expected?
- Identification performance should be better. First of all, because the short peptides bug is now fixed. But from some spot checks it seems that some PSMs also get better predictions due to the way the beams are handled. I haven't systematically verified this, so running on (part of) the 9-species benchmark to quantify the performance would be useful.

Fixes #107. Fixes #115. Supersedes #134.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+- Included the `min_peptide_len` parameter in the configuration file to restrict predictions to peptide with a minimum length.
+
 ### Changed
 
 - Calculate the amino acid scores as the average of the amino acid scores and the peptide score.
@@ -13,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 - Verify that the final predicted amino acid is the stop token.
+- The score of the stop token is taken into account when calculating the predicted peptide score.
 
 ## [3.2.0] - 2022-11-18
 
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -142,6 +142,7 @@ def main(
         max_charge=int,
         precursor_mass_tol=float,
         isotope_error_range=lambda min_max: (int(min_max[0]), int(min_max[1])),
+        min_peptide_len=int,
         dim_model=int,
         n_head=int,
         dim_feedforward=int,
--- a/casanovo/config.yaml
+++ b/casanovo/config.yaml
@@ -28,6 +28,8 @@ max_charge: 10
 precursor_mass_tol: 50  # ppm
 # Isotopes to consider when comparing predicted and observed precursor m/z's (I)
 isotope_error_range: [0, 1]
+# The minimum length of predicted peptides.
+min_peptide_len: 7
 
 # Model architecture options.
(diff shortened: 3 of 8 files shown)


melihyilmaz (commented):
Thanks Wout, I think this looks good overall. I've run this PR's version on the mouse dataset from 9-species benchmark and results are almost identical to the earlier version of Casanovo (see plot below). One issue that I realized when testing the PR was that progress bar didn't update with inference steps corresponding to batches but rather completed promptly in a few seconds at the very beginning and nothing else was printed as model finished running inference. I suspect this might be due to changes in `_forward_step()` or `on_predict_epoch_end()` but I couldn't spot what exactly it is.

I will now work on the unit tests for this PR.

PR_vs_main

I'll now cover the unit tests

melihyilmaz:
Improved unit testing for beam search decoding and all tests are passing locally. ` ModuleNotFoundError: No module named 'pytorch_lightning.lite'` does seem to be an issue for tests on github along the lines of #147. I couldn't identify the source of progress bar bug mentioned earlier, would appreciate another pair of eyes.

bittremieux:
The progress bar not properly updating is due to the generator on line 778 in `model.py`. If you change that to return a list or something it properly tracks progress again.

melihyilmaz (approved):
@bittremieux I fixed the progress bar tracking issue with a simple generator to list conversion as you suggest and all tests are passing locally. I think it's good to merge but let me know if anything's missing.

PR #139, opened 2023-02-16
Export multiple PSMs per spectrum

### LINKED ISSUE #96: Add multi-PSM output
Add option to output multiple PSMs per spectrum when running beam search. A subset or all of cached peptide predictions can be returned.

### LINKED ISSUE #138: Discard beams that include multiple N-terminal modifications
Casanovo predicts some peptides with multiple n-terminal modifications as previously highlighted in #129.

A simple solution would be to discard beams that include multiple N-terminal modifications when they're finish. I.e. just filter out the invalid predictions.

A more robust solution might be to consider N-terminal modifications as special types of stop tokens, in analogy to dots, question marks, exclamation points to end a sentence. That would require retraining the model from scratch though.

bittremieux:
- Export multiple PSMs per spectrum.
- Discard predictions with incorrect N-terminal modifications (multiple N-terminal modifications or N-terminal modifications that occur at internal positions). This is especially relevant when including multiple predictions per spectrum, as the lower-ranked PSMs are more likely to be a bit wonky.

Fixes #96. Fixes #138.

Appropriate unit tests are still missing. Additionally, first #136 should be fully merged, because these changes somewhat depends on that functionality.


melihyilmaz (changes requested):
I think everything works except for that one line I highlighted. Currently not passing the new unit tests only because of that line, can you make the change if it makes sense to you?

melihyilmaz:
As a quick sanity check I ran the PR on 9-species mouse data set and compared the predictions with results from main (before the recent merge). This PR mispredicts 6 peptides that were previously predicted correctly (out of 26k PSMs) listed below. There're also 198 peptides that they mispredict differently but no previous misprediction that this PR corrects.  I don't have a good sense why these particular peptides but I guess we don't expect the predictions to stay the same after this PR:

> PR=0 vs CN=1
> PR:  QGVHPDTGLSSK 0.9637844287432156
> CN:  VVNGAAASQPPSK 0.9393118470907212 
> 
> PR:  AGVN+0.984TVTTLVENK 0.8927332433605832
> CN:  LQLQC+57.021VVEDDK 0.8813245308895906 
> 
> PR:  DLLPDVNVFQR 0.9175506234169006
> CN:  DLNPDVNVFQR 0.8763163359835744 
> 
> PR:  YTDLLFTEQER 0.916917142768701
> CN:  SLDM+15.995SWDSVTMK 0.8911982940939757 
> 
> PR:  -17.027QGNVLLSWTDQN+0.984SK 0.7653195192106068
> CN:  -17.027QNGVLNSWTDQN+0.984SK 0.7584711946547031 
> 
> PR:  EEPLDFSNSPLAER 0.8485337277253469
> CN:  -17.027C+57.021DPLDFSNSPEALR 0.8214384562917303

melihyilmaz:
Turns out when compare this PR's predictions to the current main branch (after #153) in the same set up described in the above comment, there're no additional mispredictions by this PR and only diverging peptides between two branches are due to elimination of peptides with multiple N-terminal mods. Accordingly, discrepancy in predictions should be due to either #153 or #151

melihyilmaz:
I went back and compared predictions from each commit going back to #136, all were the same. So there's nothing wrong with this PR or #153 and #151. For some reason, predictions from #136 are different from those that I had obtained from the PR version of #136 (not sure at which commit), leading to diverging predictions I mentioned earlier [here](https://github.com/Noble-Lab/casanovo/pull/139#issuecomment-1485707374). I'm not sure why that's the case. 

I think we can merge this after you take a look at the point I raised in the beginning.

melihyilmaz (approved):
Looks good to me!

PR #141, opened 2023-03-01
Note that inference should only use a single GPU

### LINKED ISSUE #140: Multi-GPU is broken
When running inference on a multi-GPU setup ('ddp' strategy) with n GPUs, the mztab output only has 1/n results.

Unlike #128 I am able to successfully run to completion, but the output is missing results. In that issue it looks like you've had issues with multi-GPU inference in the past, but I didn't see an open bug for it.

After a little digging, it looks like it might be related to this: https://github.com/Lightning-AI/lightning/issues/10618 and the solution is to implement a different callback that gets called per process. I'm not sure of the implications on the current mztab writer strategy.

bittremieux:
Fixes #140.


wsnoble (approved):
Looks good.

PR #153, opened 2023-03-23
Migrate to pyproject.toml and restrict depthcharge version

wfondrie:
This PR accomplishes a few things:

1. It simplifies the package metadata by putting everything into pyproject.toml, eliminating the need for setup.cfg and setup.py, per [PEP621](https://peps.python.org/pep-0621/). Also see the [setuptools docs](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#) for more info.
2. It restricts the version of depthcharge, in preparation for the breaking change that is about to occur!
3. Skips our currently broken integration test.
4. Add the `test_path` directory to .gitignore.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,7 +61,6 @@ dev = [
     "psims",
     "pyteomics",
     "pytest",
-    "requests",
 ]
 
 [project.scripts]


pyproject.toml
```diff
@@ -1,12 +1,83 @@
...
+    "tqdm",
+]
+dynamic = ["version"]
+
+[project.readme]
+file = "README.md"
+content-type = "text/markdown"
+
+[project.urls]
+Homepage = "https://github.com/wfondrie/mokapot"
```
bittremieux, line 46: **todo:** Change to Casanovo rather than mokapot.

pyproject.toml
```diff
@@ -1,12 +1,83 @@
...
+    "sphinx-book-theme>=0.3.3"
+]
+dev = [
+    "pre-commit>=2.7.1",
+    "black>=19.10b0",
+    "ppx",
+    "psims",
+    "pyteomics",
+    "pytest",
+    "requests",
```
bittremieux, line 64: **nitpick:** Not really necessary here because it's a core dependency as well.

bittremieux (changes requested):

pyproject.toml
wfondrie: Whoops! Thanks.

pyproject.toml
wfondrie: Are all four of these included in another package?

wfondrie:
Thanks @melihyilmaz, I've added the test back and fixed the bug I introduced in the other PR!

casanovo/config.py
```diff
@@ -126,3 +126,7 @@ def validate_param(self, param: str, param_type: Callable):
             raise TypeError(
                 f"Incorrect type for configuration value {param}: {err}"
             )
+
+    def items(self) -> tuple[str, ...]:
```
melihyilmaz, line 130: Locally failing the unit tests, getting a `TypeError: 'type' object is not subscriptable` because of this line. Passes all tests when `-> tuple[str, ...]` commented out.

pyproject.toml
bittremieux: No, I meant that `requests` is also listed under the standard dependencies, and thus is not explicitly necessary here anymore.

casanovo/config.py
wfondrie: Ah yeah, I forgot using lowercase `tuple ` is a newer Python feature. Maybe added in 3.10?

wfondrie:
Just fixed the config file name issue as noted by @melihyilmaz.

bittremieux (approved):

PR #164, opened 2023-04-10
Fix validation step

melihyilmaz:
I realized that we hadn't updated the `validation_step()` after the recent changes to reflect the new output format of the `forward()`. Fixed this, got rid of some checks that are now performed inside `beam_search_decoding()` and added a simple unit test.


wfondrie (approved):
Looks good to me 👏

PR #165, opened 2023-04-11
Add validation frequency option 

melihyilmaz:
I put together this PR after coming to the conclusion that there was no option to run validation before a training epoch ends, which is crucial when training on large datasets like MSKB, but it also includes fixes to issues with logging and checkpointing we introduced earlier (1st commit: e11b80ada0e66f62a64fa880a33bdeb3865d99d1). To list the changes:

- `every_n_train_steps` option now controls both the frequency of validation epochs and model checkpointing. We can decouple the two but kept it as such for simplicity.
-  Validation performance metrics are logged (and added to tensorboard) at the validation epoch, and training loss is logged at the end of training epoch, i.e. training and validation metrics are logged asynchronously.
- We now log `step` rather than `epoch` as the unit of progress.
- Fixed `tb_summarywriter` not being read of the `config.yaml`.
- Fixed metrics being logged, discontinued legacy metric `recall`.
- Fixed `save_model` option, previously didn't have any effect. 

I tested the changes with different settings and things seem to work but our unit tests still don't cover the changed functions.

### DIFF
--- a/casanovo/config.yaml
+++ b/casanovo/config.yaml
@@ -81,7 +81,7 @@ residues:
   "+43.006": 43.005814      # Carbamylation
   "-17.027": -17.026549     # NH3 loss
   "+43.006-17.027": 25.980265      # Carbamylation and NH3 loss
-# Logging frequency in training epochs
+# Logging frequency in training steps
 n_log: 1
 # Tensorboard object to keep track of training metrics
 tb_summarywriter:
@@ -117,7 +117,7 @@ save_model: True
 model_save_folder_path: ""
 # Set to "False" to save the PyTorch model instance
 save_weights_only: True
-# Model checkpointing frequency in training steps
+# Model validation and checkpointing frequency in training steps
 every_n_train_steps: 50_000
 # Disable usage of a GPU (including Apple MPS):
 no_gpu: False
--- a/casanovo/config.yaml
+++ b/casanovo/config.yaml
@@ -81,7 +81,7 @@ residues:
   "+43.006": 43.005814      # Carbamylation
   "-17.027": -17.026549     # NH3 loss
   "+43.006-17.027": 25.980265      # Carbamylation and NH3 loss
-# Logging frequency in training steps
+# Logging frequency in training epochs
 n_log: 1
 # Tensorboard object to keep track of training metrics
 tb_summarywriter:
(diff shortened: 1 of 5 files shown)


casanovo/config.yaml
```diff
@@ -117,7 +117,7 @@ save_model: True
 model_save_folder_path: ""
 # Set to "False" to save the PyTorch model instance
 save_weights_only: True
-# Model checkpointing frequency in training steps
+# Model validation and checkpointing frequency in training steps
```
wfondrie, line 120: **question (non-blocking):** Do we want separate parameters for these? I haven't been training models so I don't know if this is sufficient or if they should be split.

wfondrie (approved):
This looks good to me. I left one non-blocking question though. You should also make sure to note this in the changelog.

casanovo/config.yaml
melihyilmaz: Separation would give flexibility to cases users who don't want to checkpoint as often as the validation but I think this keeps things simpler and works fine.

casanovo/denovo/model.py
```diff
@@ -890,36 +894,35 @@ def _log_history(self) -> None:
...
+                        ("eval/val_pep_precision", "valid_pep_precision"),
+                        ("eval/val_aa_precision", "valid_aa_precision"),
                     ]:
-                        self.tb_summarywriter.add_scalar(
-                            descr,
-                            metrics.get(key, np.nan),
-                            metrics["epoch"] + 1,
-                        )
+                        metric_value = metrics.get(key, np.nan)
+                        if metric_value is not np.nan:
```
bittremieux, line 920: **issue:** This is not fully correct, you should use `np.isnan` instead.

casanovo/denovo/model.py
```diff
@@ -890,36 +894,35 @@ def _log_history(self) -> None:
...
-                    "Epoch\tTrain loss\tValid loss\tAA precision\tAA recall\t"
-                    "Peptide recall"
-                )
-            metrics = self._history[-1]
-            if metrics["epoch"] % self.n_log == 0:
+        if len(self._history) == 1:
+            logger.info(
+                "Step\tTrain loss\tValid loss\tPeptide precision\tAA precision"
+            )
+        metrics = self._history[-1]
```
bittremieux, line 901: **issue:** This needs to go after making sure that the log is not empty.

PR #171, opened 2023-04-17
Only split off known extensions from output filename

bittremieux:
Old behavior:
- `--output foo.bar` -> `foo.log` and `foo.mztab`
- `--output foo.mztab` -> `foo.log` and `foo.mztab`
- `--output foo.log` -> `foo.log` and `foo.mztab`

New behavior:
- `--output foo.bar` -> `foo.bar.log` and `foo.bar.mztab`
- `--output foo.mztab` -> `foo.log` and `foo.mztab`
- `--output foo.log` -> `foo.log` and `foo.mztab`

### DIFF
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -96,7 +96,8 @@ def main(
             f"casanovo_{datetime.datetime.now().strftime('%Y%m%d%H%M%S')}",
         )
     else:
-        output = os.path.splitext(os.path.abspath(output))[0]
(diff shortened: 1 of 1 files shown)


wsnoble (commented):
What happens if the input is "foo" (I.e., there is no extension)?

Otherwise, this looks good.

melihyilmaz:
> What happens if the input is "foo" (I.e., there is no extension)?
> 
> Otherwise, this looks good.

It'd be:
--output foo -> foo.log and foo.mztab

Issue #157, opened 2023-03-29
Replace the `--mode` argument with subcommands


wfondrie:
Currently, casanovo relies on the required `--mode` argument to know what to do. I propose replacing this with subcommands so that:

- `casanovo --mode denovo ...` becomes `casanovo run ...`
- `casanovo --mode train ...` becomes `casanovo train ...`
- `casanovo --mode eval ...` becomes `casanovo evaluate ...`

Please note that this would be a *breaking change.*

bittremieux:
Fixed in #184.

PR #184, opened 2023-05-11
CLI revamp

wfondrie:
This PR updates the CLI to take more canonical, subcommand form (#157) and updates the documentation according. 

It also ups the number of epochs used for our integration test.


wfondrie:
I've configured a new GH Action workflow to update any terminal screenshots we have in the docs, automagically! 🪄 


Screenshot 2023-05-11 at 12 45 05 AM

wfondrie:
@bittremieux, I also went ahead and added the option to calculate the precision metrics during training and restricted the depthcharge version.

casanovo/casanovo.py
```diff
@@ -7,116 +7,294 @@
...
+@main.command()
+def version() -> None:
+    """Get the Casanovo version information"""
+    versions = [
+        f"Casanovo: {__version__}",
+        f"Depthcharge: {depthcharge.__version__}",
+        f"Lightning: {lightning.__version__}",
+        f"PyTorch: {torch.__version__}",
+    ]
+    sys.stderr.write("\n".join(versions) + "\n")
```
bittremieux, line 217: **question:** Why is this written to stderr instead of using the logger?

casanovo/casanovo.py
```diff
@@ -7,116 +7,294 @@
...
 )
-def main(
-    mode: str,
+def configure(output: str) -> None:
+    """Generate a Casanovo configuration file to customize.
+
+    The casanovo configuration file is in the YAML format.
+    """
+    Config.copy_default(output)
+    sys.stderr.write(f"Wrote {output}\n")
```
bittremieux, line 234: **question:** Same question here, why not use the logger?

casanovo/denovo/model.py
```diff
@@ -759,19 +764,26 @@ def validation_step(
...
-        )
-        log_args = dict(on_step=False, on_epoch=True, sync_dist=True)
-        self.log(
-            "Peptide precision at coverage=1",
-            pep_precision,
-            **log_args,
-        )
-        self.log("AA precision at coverage=1", aa_precision, **log_args)
 
+        if self.calculate_precision:
```
bittremieux, line 768: **question:** This only doesn't calculate the metrics, but peptide inference is already done in the preceding lines I think? The latter is the very expensive step, the former adds only a small amount of extra runtime.

bittremieux (commented):
This looks pretty good. The new CLI will be more user-friendly.

One thing I'm not immediately certain about is whether the changes related to `calculate_precision` really result in a significant speed-up? It seems to me that only the AA and peptide metrics are not computed when this is set to False. Although calculating those metrics does take a bit of time, this is not excessive. The main slow-down is due to peptide inference. So I think we should make peptide inference optional as well during validation, and only recording the loss when `calculate_precision` is False. This should give us the biggest runtime improvement during validation.

wfondrie:
Your absolutely right - I made a mistake and intended to put the `if` prior to decoding 🤦‍♂️

casanovo/casanovo.py
wfondrie: Yikes! I don't remember doing that, but I should fix it 😛

casanovo/denovo/model.py
wfondrie: You're right - it should be fixed!

casanovo/casanovo.py
wfondrie: The main reason is that this command is isolated and occurs outside of setting up logging in the `setup()` function. I guess the other question is should this actually go to stdout, since it is the expected output of the command?

bittremieux: Ok, makes sense. Although maybe stdout is indeed more appropriate then?

casanovo/casanovo.py
bittremieux: Yes, maybe we can indeed use the logging, and separate the setup into (i) initializing the logging, and (ii) initializing the model? Then we can consistently use (i) for all commands.

bittremieux:
@wfondrie I made small changes to the logging for `version` and `configure`, as discussed during the previous code review. Please have a look at the changes, and if they seem fine to you, we can merge this PR.

wfondrie:
Woohoo all test pass! Merge with prior permission from @bittremieux.

PR #195, opened 2023-06-19
Use a single beam by default

bittremieux:
And ensure that there are correct results when using only a single beam.

@melihyilmaz: Please include your unit tests for only a single beam on this branch.


melihyilmaz (approved):
I added unit test to cover the previously buggy functionality and ran inference locally to confirm single beam decoding works as intended.

PR #197, opened 2023-06-20
Run linting and tests for the dev branch

bittremieux:


bittremieux:
Hmm, apparently the test still fails, due to a stop token being included when calculating the peptide mass. Considering that this PR doesn't change any of the Casanovo code and the previous commit on `dev` passed all tests, is this possibly due to a change that meanwhile happened in DepthCharge @wfondrie (despite pinning to version 0.2.x)?

PR #200, opened 2023-06-27
Fix linting errors

bittremieux:


bittremieux:
There does seem to be an issue with failing tests on `dev` that we should fix. I can't immediately figure out what's going wrong, so this warrants a closer look.

melihyilmaz:
Locally (w/ or w/o GPU) there's a failure in an earlier assertion as well but when I commented that out I was able to reproduce the same failure. I also don't have a good intuition about what might be wrong but I'll take a closer look. @wfondrie might chime in as well since I think he had written the integration test?
image

bittremieux:
Fixed in #184.

Issue #203, opened 2023-07-02
Tensorboard is not working


wsnoble:
When I train a model with tensorboard turned on, I get a numpy error about mismatched dimensions:

```
casanovo --mode=train --config config.yaml --peak_path=target.train.100k.mgf --peak_path_val=target.train.100k.mgf
...
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.
```

The full stderr output, as well as the input file and the config file, are attached.  This same command works fine if I turn off tensorboard output in the config file.

[stderr.txt](https://github.com/Noble-Lab/casanovo/files/11931615/stderr.txt)
[config.txt](https://github.com/Noble-Lab/casanovo/files/11931616/config.txt)
[target.train.100k.mgf.txt](https://github.com/Noble-Lab/casanovo/files/11931617/target.train.100k.mgf.txt)

bittremieux:
Can you try to change `metric_value` in [this line](https://github.com/Noble-Lab/casanovo/blob/26b08cf4ba6f09bd64a2bbddb8c1975788cdbe4e/casanovo/denovo/model.py#L918C36-L918C48) to `metric_value.cpu()` and see if it works then?

wsnoble:
I tried this:

```
                   metric_value = metrics.get(key, np.nan)
                    if not np.isnan(metric_value):
                        self.tb_summarywriter.add_scalar(
                            descr, metric_value.cpu(), metrics["step"]
                        )
```

The behavior remains unchanged.

cfmelend:
Was able to replicate this bug with the same solution working for me.

bittremieux:
Apparently it works on `dev`, closing.

PR #209, opened 2023-07-14
Tb fix

cfmelend:
A quick fix to the tb TypeError issue seen [here](https://github.com/Noble-Lab/casanovo/issues/203). It seems like the `if not np.isnan(metric_value)` call in [this line](https://github.com/Noble-Lab/casanovo/blob/26b08cf4ba6f09bd64a2bbddb8c1975788cdbe4e/casanovo/denovo/model.py#L916C3-L916C3) is trying to convert some of the tensor values in metric_value (e.g. the train_loss) to numpy values. From what I can tell some of the values in metrics are floats and some are tensors so I added a short check to see if there are any tensors and if so make a copy of the variable, send to cpu, send to numpy (the last part might be wholly unnecessary).

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -914,14 +914,13 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
                 ]:
                     metric_value = metrics.get(key, np.nan)
                     if metric_value is not None:
-                        if torch.is_tensor(
-                            metric_value
-                        ):  # Check if metric_value is a torch tensor
+                        # Check if metric_value is a torch tensor
+                        if torch.is_tensor(metric_value):
                             metric_value_np = (
                                 metric_value.clone().cpu().numpy()
-                            )  # tensor losses
+                            )
                         else:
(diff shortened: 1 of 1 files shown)


casanovo/denovo/model.py
```diff
@@ -913,10 +913,18 @@ def _log_history(self) -> None:
                     ("eval/val_aa_precision", "valid_aa_precision"),
                 ]:
                     metric_value = metrics.get(key, np.nan)
-                    if not np.isnan(metric_value):
-                        self.tb_summarywriter.add_scalar(
-                            descr, metric_value, metrics["step"]
-                        )
+                    if metric_value is not None:
```
bittremieux, line 916: **suggestion:** This seems a bit overkill. Does it also work by doing `new_style=True` in [`set_scalar(...)`](https://pytorch.org/docs/stable/tensorboard.html#torch.utils.tensorboard.writer.SummaryWriter.add_scalar)?

cfmelend: This works but it seemed to suppress logging output for the val/train loss. Tensorboard seems to work correctly on dev branch.

PR #220, opened 2023-07-28
Remove Pytorch Lightning logger

### LINKED ISSUE #219: Remove confusing "logger" option in config
The logger option in the config file is almost completely unused. It can apparently be used to enable/disable logging in the Pytorch Lightning Trainer.  I don't really understand what this means from a practical perspective, but it seems like we should get rid of this option entirely.

bittremieux:
Fixes #219.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,16 +6,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
-### Changed
-
-- The CLI has been overhauled to use subcommands.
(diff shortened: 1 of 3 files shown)


wsnoble (commented):
What does the logger actually do? I'm just wondering if we should set it to something every time.

bittremieux:
It enables you to log to Tensorboard or alternative loggers (e.g. CSV file) directly through Pytorch Lightning, but we do this externally to track extra metrics (AA and peptide precision).

VarunAnanth2003 (approved):
Looks good to me, I don't see any issues arising from these changes and all tests pass.

PR #224, opened 2023-07-31
Fix validation and checkpointing interval

### LINKED ISSUE #211: Problematic check_val_every_n_epoch message
I got this error message from Casanovo:

```
ValueError: `val_check_interval` (29) must be less than or equal to the number of the training batches (1). If you want to disable validation set `limit_val_batches` to 0.0 instead.If you want to validate based on the total training batches, set `check_val_every_n_epoch=None`.
```

The problem is that I can't see any way to change `check_val_every_n_epoch` in the config file.  There is something called `every_n_train_steps`, but if I set this to `None`, then I get

```
TypeError: Incorrect type for configuration value every_n_train_steps: invalid literal for int() with base 10: 'None'
```

bittremieux:
The option to configure the interval for validation and checkpointing has been changed:
- Config option `every_n_train_steps` has been renamed to `val_check_interval` in accordance to the corresponding Pytorch Lightning parameter.
- This now determines the number of training steps _across_ epochs instead of _within_ a single epoch. I.e. previously if there were 75,000 batches in the training data, validation would run at epoch=0,steps=50k; epoch=1,steps=50k; epoch=2,steps=50k; ...; whereas now it will run at epoch=0,steps=50k; epoch=1,steps=25k; epoch=1,steps=75k; ...

Fixes #211.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - The CLI has been overhauled to use subcommands.
 - Upgraded to Lightning >=2.0
 - Checkpointing is now configured to save the top-k models instead of all.
+- `every_n_train_steps` has been renamed to `val_check_interval` in accordance to the corresponding Pytorch Lightning parameter.
 
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


VarunAnanth2003 (approved):
CodeCov coverage did not change and all tests passed. Looked at code and no changes will break anything. Good to go

PR #229, opened 2023-08-07
Fix custom residues in config

### LINKED ISSUE #227: Changing Modifications in configuration file
Dear Authors, 

thank you very much for providing casanovo. I was trying to train new casanovo model using the Human HCD spectral library (containing 1,114,503 peptides). I split the data set into 99:1. I changed the modifications in the configuration file of casanovo to use only C and M (Oxidation), because I am interested only in identifying these modifications. The training data set contains sequences which also have Q and N and some others. But as I am interested in only two modifications assigned in the configuration file of casanovo, this should be no problem correct? 
When I started the training process, casanovo showed these modifications as input modifications:

`residues = {'G': 57.021464, 'A': 71.037114, 'S': 87.032028, 'P': 97.052764, 'V': 99.068414, 'T': 101.04767, 'C+57.021': 160.030649, 'L': 113.084064, 'I': 113.084064, 'N': 114.042927, 'D': 115.026943, 'Q': 128.058578, 'K': 128.094963, 'E': 129.042593, 'M': 131.040485, 'H': 137.058912, 'F': 147.068414, 'R': 156.101111, 'Y': 163.063329, 'W': 186.079313, 'M+15.995': 147.0354, 'N+0.984': 115.026943, 'Q+0.984': 129.042594, '+42.011': 42.010565, '+43.006': 43.005814, '-17.027': -17.026549, '+43.006-17.027': 25.980265}`  

Attached is my configuration file ([config.yml.txt](https://github.com/Noble-Lab/casanovo/files/12267464/config.yml.txt))
This is also the log file of the training start:  [log.txt](https://github.com/Noble-Lab/casanovo/files/12267478/log.txt)

After finishing the training, I tested the model with the used configuration file and unfortunately other modifications are identified! 

Should I remove all peptides from the training set, which contain other modifications?! I though that identifying the modifications in the configuration file should be the input for the model. 

Thank you very much for your support! 
Best,

Ahmad

### LINKED ISSUE #228: User's custom residue vocab not loaded
It looks like the validation logic in `config.py` never actually overwrites the `self._params["residues"]` with the value in `self._user_config`. Specifically, [validate_param](https://github.com/Noble-Lab/casanovo/blob/fba790d0abe3f9fad1041ed4185ff90043af8c2f/casanovo/config.py#L87C27-L87C27) is never called for "residues".

Quick fix is to add `residues=dict,` to the [`_config_types` dictionary](https://github.com/Noble-Lab/casanovo/blob/fba790d0abe3f9fad1041ed4185ff90043af8c2f/casanovo/config.py#L35C16-L35C16). The value (`dict`) will be ignored because the "residues" key is special-cased in the validation, but that should be fine (I think the exception will be triggered if it's not a dict).

Cause of #227.

bittremieux:
Custom residues in the config file were ignored, making it impossible to retrain Casanovo with now PTMs for example.

Fixes #228, fixes #227.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Casanovo now runs on CPU and can passes all tests.
 - Upgrade to Depthcharge v0.2.0 to fix sinusoidal encoding.
 - Correctly refer to input peak files by their full file path.
+- Specifying custom residues to retrain Casanovo is now possible.
 
 ## [3.3.0] - 2023-04-04
 
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -50,6 +50,7 @@ class Config:
         dropout=float,
         dim_intensity=int,
         max_length=int,
+        residues=dict,
         n_log=int,
         tb_summarywriter=str,
         warmup_iters=int,
@@ -66,9 +67,9 @@ class Config:
         save_top_k=int,
         model_save_folder_path=str,
         val_check_interval=int,
+        calculate_precision=bool,
         accelerator=str,
         devices=int,
-        calculate_precision=bool,
     )
 
     def __init__(self, config_file: Optional[str] = None):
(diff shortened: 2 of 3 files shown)


wfondrie (approved):
LGTM. Sorry for the delay!

PR #230, opened 2023-08-07
Upgrade to pytorch-lightning 2

AmirAlavi:
This migrates the code to the latest major version of pytorch-lightning. I tried to do it with minimal changes to stay true to the original design philosophy of Casanovo. Feedback is much appreciated.

Summary of changes (I followed [this guide](https://lightning.ai/docs/pytorch/latest/upgrade/from_1_9.html) to upgrade):

- LightningLite is deprecated
  - Switch to Fabric
- `auto_select_gpus` parameter for `Trainer` is deprecated
  - Use `find_usable_cuda_devices` in `_get_devices` instead. Also, `devices` can't be 0, so now `_get_devices` falls back to `auto`
- Relying on passed in `outputs` in `on_predict_epoch_end` is deprecated
  - Use `self.trainer.predict_loop.predictions` instead
- Passing a dictionary to `LightingModule.log` is deprecated
  - Log as scalar with the "mode" (e.g. "train" or "valid") embedded in string key to the same effect
- `strategy` parameter for `Trainer` cannot be `None`
  - Return "auto" if no cuda devices

### DIFF
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -50,6 +50,7 @@ class Config:
         dropout=float,
         dim_intensity=int,
         max_length=int,
+        residues=dict,  # note, this key is special-cased and type is ignored
         n_log=int,
         tb_summarywriter=str,
         warmup_iters=int,
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -15,10 +15,8 @@ import appdirs
 import click
 import github
 import requests
-import torch
 import tqdm
-import yaml
-from pytorch_lightning.lite import LightningLite
+from lightning import Fabric
 
 from . import __version__
 from . import utils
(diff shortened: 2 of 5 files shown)


casanovo/config.py
```diff
@@ -50,6 +50,7 @@ class Config:
         dropout=float,
         dim_intensity=int,
         max_length=int,
+        residues=dict,  # note, this key is special-cased and type is ignored
```
AmirAlavi, line 53: This is already fixed by #229 so I guess I should remove here?

wfondrie:
Hi @AmirAlavi - thank you for your PR, but we've actually already begun this migration and it will be present in the next release. You can check out the changes in #176.

Casanovo uses a Git Flow-type of system, where our in-development work goes into the `dev` branch before being migrated into main. There's a lot of other updates waiting to be merged into main too!

AmirAlavi:
Ah got it, thanks @wfondrie, don't know how I missed that PR. Thanks! Will stay tuned.

wfondrie:
No worries - we need to update our contribution docs to be more accurate since the change. Sorry for the confusion!

PR #234, opened 2023-08-17
Added shuffling to trainset dataloaders, no shuffle by default for ot…

cfmelend:
By default casanovo does not shuffle the dataloaders' batches. This quick patch simply adds an arg in dataloaders.py, within the `DeNovoDataModule` class, and within the `_make_loader()` method that determines shuffling. This param is set to false by default and set to true only when `_make_loader()` is called within the `train_dataloader()` method.


wfondrie (approved):
Looks good to me!

Issue #245, opened 2023-09-21
precursor_mass_tol is being ignored


wsnoble:
I can't seem to change the behavior of the precursor mass filter, even if I set the precursor_mass_tol to a ridiculously large value.  Here is a small example where I set it to 1e10, and there are still negative scores.

We should also provide (or document, if it exists) a way to turn off the filter entirely.

Also, separately, it seems like the auto-download of the model only works for the `train` command.  When I try it with `sequence` I get `ValueError: A model file must be provided`.

[sample.mztab.txt](https://github.com/Noble-Lab/casanovo/files/12684785/sample.mztab.txt)
[sample.mgf.txt](https://github.com/Noble-Lab/casanovo/files/12684786/sample.mgf.txt)
[sample.log.txt](https://github.com/Noble-Lab/casanovo/files/12684787/sample.log.txt)
[casanovo.yaml.txt](https://github.com/Noble-Lab/casanovo/files/12684788/casanovo.yaml.txt)

melihyilmaz:
Opened a PR to fix the two bugs but functionality to turn off filter entirely currently doesn't exist and it might be non-trivial to implement since it's intertwined with decoding procedure so I'm opening a new enhancement issue to cover that later.

PR #246, opened 2023-09-22
Use config options and auto-downloaded weights

melihyilmaz:
Fixes two issues raised by  #245:

- File path to auto-downloaded weights in cache currently isn't provided `ModelRunner` when initializing a model. Fixed this by changing the output of `setup_model` include it.  
- Options in config file are currently disregarded when loading a model checkpoint, fixed this and also added the error message to cover cases when model architecture related parameters are mismatched between the checkpoint and config file.


wfondrie:
> Options in config file are currently disregarded when loading a model checkpoint, fixed this and also added the error message to cover cases when model architecture related parameters are mismatched between the checkpoint and config file.

This was to some degree intentional, so that a user did not have to know exactly the parameters that were used to initialize a model in order to load the weights. When was this causing an issue?

casanovo/casanovo.py
```diff
@@ -378,7 +378,7 @@ def setup_model(
     for key, value in config.items():
         logger.debug("%s = %s", str(key), str(value))
 
-    return config
+    return config, model
```
wfondrie, line 381: **question:** Why does the model need to be returned?

wfondrie (commented):
Can you elaborate more on what this PR does and the reasoning for the changes?

casanovo/casanovo.py
melihyilmaz: Because the automatically downloaded weights aren't used otherwise - they're returned with `_get_model_weights()` inside the function.

wfondrie:
> I think having the default config file reflect the parameters for the model for which there're released weights would be enough - as long as users don't change those they'd be good and they can still modify other parameters, they'd get an error message if they deviate from config defaults while using our weights.

I don't think this is a very robust, particularly in the case when users may want to change models (for example, to a non-tryptic model). Are we always to going to re-train all of our submodels with an identical architecture to the main model? 

 > Users might still want to modify non-architecture related parameters, like precursor_mass_tol in https://github.com/Noble-Lab/casanovo/issues/245,
 
 I think the better solution would be to explicitly support overwriting these parameters of model weights, maybe including a warning that the loaded weights do not match the parameter file.

wsnoble:
I agree with Will that it would be better to use the parameters associated with the given model and to issue a warning to the user that the parameters in the provided config file do not match the given model (explicitly listing the given parameter value and the value that was used).

wfondrie (approved):
Looks good to me. I fixed the screenshot action that was failing.

PR #249, opened 2023-09-28
Add option to change learning rate scheduler

justin-a-sanders:
Added config option to choose between the current cosine learning rate schedule, a linearly decreasing schedule, or a no schedule (a constant learning rate). Also made it easier to add any new learning rate schedule options in the future by separating out the linear warmup into a separate schedule.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 
 - The CLI has been overhauled to use subcommands.
-- Upgraded to Lightning >=2.0
+- Upgraded to Lightning >=2.0.
 - Checkpointing is configured to save the top-k models instead of all.
 - Log steps rather than epochs as units of progress during training.
 - Validation performance metrics are logged (and added to tensorboard) at the validation epoch, and training loss is logged at the end of training epoch, i.e. training and validation metrics are logged asynchronously.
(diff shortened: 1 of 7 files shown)


bittremieux:
Can you properly format the files using black (I see single quotes) so that that test passes? The unit test failure is weird, some infinite loop when installing dependencies. 🤔

melihyilmaz:
When I locally tested the new branch and track learning rate with `print(self.lr_schedulers().state_dict())`, I saw that in `cosine` and `linear` options decay schedules did override the warm up period, i.e. no warm up, and linear warm up wasn't even linear. `constant` option works as intended.

melihyilmaz:
I created [this](https://colab.research.google.com/drive/1jFd5iqVgAggriQfh0qOeX3EPshsgRMu9?usp=sharing) Colab notebook to experiment with the implementation in this PR by plotting the learning rate over training steps from a toy example.

Cosine scheduler implementation in this PR doesn't have a warm up despite being chained with linear warm up (compare the with `main` branch version) and I couldn't figure out why yet. @justin-a-sanders it'd be great if you can take a look at the notebook.

melihyilmaz:
While tinkering with the notebook, I got the warm up to work with using our cosine scheduler using `torch.optim.lr_scheduler.SequentialLR(optimizer, schedulers = lr_schedulers, milestones = [warmup_iters])` instead of `torch.optim.lr_scheduler.ChainedScheduler(lr_schedulers)` for the `final_scheduler` but the periodicity seems to be off now (see the last cell in the notebook).

justin-a-sanders:
> While tinkering with the notebook, I got the warm up to work with using our cosine scheduler using `torch.optim.lr_scheduler.SequentialLR(optimizer, schedulers = lr_schedulers, milestones = [warmup_iters])` instead of `torch.optim.lr_scheduler.ChainedScheduler(lr_schedulers)` for the `final_scheduler` but the periodicity seems to be off now (see the last cell in the notebook).

I took a quick look at this - to be used in a chained learning rate schedule, the LR needs to be 'chainable', meaning it is a function which depends on only the current learning rate and the epoch (and not some initial value). Unless I'm missing something, I don't think a cosine lr schedule is easily compatible with that paradigm. I put a hacky example of how you can make it work in the notebook by dividing by the previous factor each epoch, but this either breaks after one cycle (div by 0) or requires you to slightly shift the cosine so it is never exactly zero. I'm not sure the flexibility of easily swapping in different schedules is worth the hack if you have already found warmup+cosine gives best results.

bittremieux:
Superseded by #294.

PR #253, opened 2023-10-13
Correctly report AA precision and recall during validation

### LINKED ISSUE #252: Bug in evaluation script during training
There are some inconsistencies in how arguments are passed to the various evaluation functions. This might be causing amino acid precision and recall to be switched.

In the [validation step](https://github.com/Noble-Lab/casanovo/blob/ea2e8751824cfc31ecb40d2dd44a499905e8805c/casanovo/denovo/model.py#L761), the peptide precision is calculated as follows:
```python
        aa_precision, _, pep_precision = evaluate.aa_match_metrics(
            *evaluate.aa_match_batch(
                peptides_pred, peptides_true, self.decoder._peptide_mass.masses
            )
        )
```

The [evaluate.aa_match_metrics](https://github.com/Noble-Lab/casanovo/blob/ea2e8751824cfc31ecb40d2dd44a499905e8805c/casanovo/denovo/evaluate.py#L243) expects the arguments in the order of first `n_aa_true` and then `n_aa_pred`.

```python
def aa_match_metrics(
    aa_matches_batch: List[Tuple[np.ndarray, bool]],
    n_aa_true: int,
    n_aa_pred: int,
) -> Tuple[float, float, float]:
```

The [evaluate.aa_match_batch](https://github.com/Noble-Lab/casanovo/blob/ea2e8751824cfc31ecb40d2dd44a499905e8805c/casanovo/denovo/evaluate.py#L184C1-L191C53) function returns `n_aa1` and `n_aa2` in the same order as `peptides1` and `peptides2`

```python
def aa_match_batch(
    peptides1: Iterable,
    peptides2: Iterable,
    aa_dict: Dict[str, float],
    cum_mass_threshold: float = 0.5,
    ind_mass_threshold: float = 0.1,
    mode: str = "best",
) -> Tuple[List[Tuple[np.ndarray, bool]], int, int]:
    ...
    return aa_matches_batch, n_aa1, n_aa2
```

The problem comes in that in the validation step, we pass `peptides1=peptides_pred` and `peptides2=peptides_true`, thereby passing `n_aa_true` as the number of number of predicted amino acids, and vice-versa for `n_aa_pred`.

This shouldn't cause an issue to the current peptide-level precision calculation, but it may affect older versions of the codebase.

bittremieux:
Fixes #252.

AA precision and recall were swapped due to swapped input of true and predicted peptides to the evaluation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Casanovo runs on CPU and can passes all tests.
 - Enable gradients during prediction and validation to avoid NaNs from occuring as a temporary workaround until a new Pytorch version is available.
 - Upgrade to depthcharge v0.2.3 for `PeptideTransformerDecoder` hotfix.
+- Correctly report amino acid precision and recall during validation.
 
 ### Removed
 
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -773,8 +773,8 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
 
         aa_precision, _, pep_precision = evaluate.aa_match_metrics(
             *evaluate.aa_match_batch(
-                peptides_pred,
                 peptides_true,
+                peptides_pred,
                 self.decoder._peptide_mass.masses,
             )
         )


melihyilmaz (approved):
Looks good!

PR #257, opened 2023-10-24
Issue error for unrecognized/missing config file entry

ishagokhale:
added unit tests to check is exceptions were raised


melihyilmaz (approved):
Looks good to me!

PR #261, opened 2023-11-01
Label smoothing in training

melihyilmaz:
After encountering the NaN outputs from the model half way through training in a few runs, I experimented with minimal label smoothing when calculating the training loss as a mitigation strategy. I was able redo the same training runs with the same setup significantly longer without encountering NaNs and with similar performance metrics compared to the original runs.

Loss calculation is only impacted by smoothing during training steps, i.e. not validation, and I tentatively added the minimal label smoothing factor as the default option.


wfondrie (changes requested):
This looks good to me aside from needing a unit test.

wfondrie (approved):
This looks good to me!

PR #265, opened 2023-11-10
Fix spectra skipping and add min_n_peaks option

### LINKED ISSUE #264: Add min_peaks config option
Casanovo should have a configuration option that specifies the minimum number of peaks that are required to be in a spectrum before it will make a prediction.  Currently, Casanovo will happily make a prediction (sometimes with a high score) even if the spectrum has only a handful of peaks.  The default value for the analogous parameter in Tide is 20 and in Comet is 10. For de novo, I think we should use a default of 20.  This is the minimum in the 9-species benchmark.

Note that, ideally, Casanovo would track and then report something like "Info: XXX spectra skipped because they have fewer than XX peaks."

melihyilmaz:
Skips spectra with fewer than min_n_peaks (min_n_peaks=20 by default) and adds the corresponding config option. Fixes #264 

This solution is a bit hacky and discards spectra at the batching stage but I think the existing functionality for skipping spectra was buggy to begin with: it only replaces the spectra to be skipped with dummy spectra and runs the model on those dummies so that we still get predictions for them. Am I missing something, @bittremieux @wfondrie?

TODOs:
- Find a way to skip spectra before batching, e.g. inside the dataset
- Log how many spectra are skipped
- Unit tests

### DIFF
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -34,6 +34,7 @@ class Config:
     _default_config = Path(__file__).parent / "config.yaml"
     _config_types = dict(
         random_seed=int,
+        min_n_peaks=int,
         n_peaks=int,
         min_mz=float,
         max_mz=float,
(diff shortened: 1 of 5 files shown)


bittremieux:
Superseded by #455.

PR #269, opened 2023-12-06
Optimize beam search

bittremieux:
- Move some initialization code for discarding invalid beams (i.e. which tokens are neutral losses or N-terminal residues) to `__init__` to only do it when initializing the model instead of every time a batch of beams is checked for termination.
- Move peptide detokenizing to the point where it's absolutely needed. This is still a slow step (currently done in DepthCharge) that could maybe be skipped by calculating masses from tokens directly? @wfondrie What do you think?
- Only calculate the peptide mass once and then subtract masses of neutral losses instead of adding the neutral losses to the peptide and then calculating the mass of the entire peptide.
- Early stopping when comparing delta masses for different isotopes.
- Numba JIT compilation of mass error calculation. Although this requires moving the precursor m/z and charge tensors to CPU, so I don't know whether it's actually beneficial. Any thoughts @wfondrie @melihyilmaz?

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -324,10 +324,9 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
         ----------
         tokens : torch.Tensor of shape (n_spectra * n_beams, max_length)
             Predicted amino acid tokens for all beams and all spectra.
-         scores : torch.Tensor of shape
-         (n_spectra *  n_beams, max_length, n_amino_acids)
-            Scores for the predicted amino acid tokens for all beams and all
-            spectra.
+        precursors : torch.Tensor of size (n_spectra * n_beams, 3)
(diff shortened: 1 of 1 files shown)


wfondrie:
> Move peptide detokenizing to the point where it's absolutely needed. This is still a slow step (currently done in DepthCharge) that could maybe be skipped by calculating masses from tokens directly? @wfondrie What do you think?

I'm working on a bigger PR for depthcharge right now and can add support for this 👍 

> Numba JIT compilation of mass error calculation. Although this requires moving the precursor m/z and charge tensors to CPU, so I don't know whether it's actually beneficial. Any thoughts @wfondrie @melihyilmaz?

That's a good question. I don't know either 🤔. I guess the alternative is to move the calculated m/z to the GPU and perform calculations using pytorch.

PR #275, opened 2023-12-25
Remove `train_from_scratch` config option

### LINKED ISSUE #263: Remove train_from_scratch
We should delete the train_from_scratch option from the config file.  If the user provides a model using the -m option, then we should start from that.  Otherwise, we train from scratch.  With the current setup, if a user provides a model but forgets to change this Boolean then the given model is ignored.

bittremieux:
Instead of having to specify `train_from_scratch` in the config file, training will proceed from an existing model weights file if this is given as an argument to `casanovo train`.

Fixes #263.


melihyilmaz (approved):
LGTM, thanks Wout!

PR #276, opened 2023-12-25
Update `output` command-line description

### LINKED ISSUE #241: fix documentation of --output
In the command line documentation, the --output option is only for the sequencing phase and doesn't reflect what outputs are produced during training.

```
│    --output               -o  FILE                    The mzTab file to      │
│                                                       which results will be  │
│                                                       written.               │
```

bittremieux:
Fixes #241.


wsnoble:
How does Casanovo handle model file naming when fine-tuning a model?  In this case, it seems like --model will be used to specify the name of the input file AND the output file. Problem!  Do we need separate --model-in and --model-out options? Alternatively, is there some scheme whereby we name the trained model with the epoch number in the filename? If so, this should be documented here.

melihyilmaz:
> How does Casanovo handle model file naming when fine-tuning a model?

By default, checkpoint files are named in the `epoch=40-step=2550000.ckpt` format regardless of the input ckpt filename. This is the case for both fine-tuning and training from scratch.

melihyilmaz:
> How does Casanovo handle model file naming when fine-tuning a model?

By default, checkpoint files are named in the `epoch=40-step=2550000.ckpt` format regardless of the input ckpt filename. This is the case for both fine-tuning and training from scratch.

wsnoble:
OK, so this is obviously bad, because if I train two models from the same working directory, the files will overwrite one another.  We should change this to be `<model-name>.epoch=40.step=2500000.ckpt`, where <model-name> comes from the --model parameter.  This behavior should also be documented.

wsnoble:
I just ran into another problem which needs to be fixed, probably on this PR.  It seems like Casanovo is trying to do something smart if I specify an `--output` option with a period in it.  Anyway, the result is that if I do something like `--output foo.bar`, then the output is `foo.log`.  Worse still, it seems like we have some logic in there that prevents it from overwriting existing files, but it does this silently.  So I ran a big sequencing run using a name like "foo.filtered" but because "foo.mztab" and "foo.log" already existed, Casanovo simply ran to completion but did not produce any output files.

As we discussed, we should add an `--overwrite` flag that allows users to ovewrite existing files. Otherwise, Casanovo should halt with an error if it tries to overwrite an existing file. Ideally, when sequencing it should check for file existence before starting the run, rather than at the end.

The argument to `--output` should be used as the prefix for all output files, with no attempt to parse it.  The argument may contain directory names, but we will make no attempt to verify that they exist.

bittremieux:
Superseded by #372.

PR #280, opened 2024-01-20
Use geometric mean to calculate peptide score

bittremieux:


melihyilmaz:
I compared arithmetic vs. geometric mean based peptide scores at inference time for both n_beams=1 and n_beams=10 on the mouse dataset from 9-species (where we have the lowest AP for Casanovo). These results are inline with the earlier results: even when we use geometric beam at inference time to select the highest scoring peptide for n_beams=10 we predict the exact same peptide as arithmetic mean ~90% of the time. In the remaining 10% of all spectra where we predict a different peptide with geometric mean, we’re mostly switching from one precursor filter-fitting incorrect peptide to another filter-fitting incorrect peptide for the most part.

Issue #281, opened 2024-01-24
Allow for flexible GPU selection


gdewael:
Hi!

The `config.yaml` only allows the "device" to be set to an `int`, (according to https://github.com/Noble-Lab/casanovo/blob/fb6fef2d7b63001a36b199fe2b8595ee69cf17ee/casanovo/config.py#L73).

In PyTorch Lightning, it is possible to specify which GPU you want to use via a list (e.g. [3, 5] to use the third and fifth GPUs, or even [6] to select 1 specifically).

Perhaps `validate_param` can be rewritten to handle the `device` parameter separately?

gdewael:
I've added [a pull request](https://github.com/Noble-Lab/casanovo/pull/282) to address this issue.

bittremieux:
A workaround is to set `CUDA_VISIBLE_DEVICES` to the GPU identifier(s) that you want to use, and then allow Casanovo to use all (visible) devices.

bittremieux:
Fixed in 6ee6351. (Accidentally pushed directly to `main`. 😳)

Instructions on how to set `devices` in the config file in combination with `CUDA_VISIBLE_DEVICES` are now included in the [FAQ](https://casanovo.readthedocs.io/en/latest/faq.html#gpu-troubleshooting).

PR #282, opened 2024-01-24
devices as a special parameter to check types for

gdewael:
Added `"devices"` as a special parameter to check the type against in `validate_param()`.

`isinstance()` allows to check the type using either a type or a tuple of types, allowing more flexibility.

In principle, every parameter could be checked using `isinstance`, although the logic is a bit different. The code for the other parameters now checks if said variable _can_ be casted to the desired type, whereas `isinstance` solely checks if they're already the correct type. (For example, specifying `devices: "2"` in the `config.yaml` file will now raise an error, whereas this was not the case previously)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -14,9 +14,9 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Setup Python 3.10
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: "3.10"
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,14 @@
 name: Lint
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main
+      - dev
+  pull_request:
+    branches:
+      - main
+      - dev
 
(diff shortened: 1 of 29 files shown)


bittremieux:
A disadvantage of this change is that if `devices` is a list, the elements of this list are not type checked. So if a user specifies `devices: ["foo", "bar"]` this will pass the config validation but result in an error further down in the code.

Maybe an alternative solution could be to use the built-in Python type hints with a library like [typeguard](https://github.com/agronholm/typeguard) to validate arbitrarily complex values?

@wfondrie what do you think?

bittremieux:
Alternatively, fine-grained GPU control can always be performed outside of Casanovo by setting [`CUDA_VISIBLE_DEVICES`](https://github.com/Noble-Lab/casanovo/issues/281#issuecomment-1910564013).

gdewael:
An alternative requiring fewer code changes, but still allowing more flexibility in the config is to change the required datatype of `devices` as `str`. In PyTorch Lightning, this still allows flexible choice of GPUs (e.g. `"2"` select whatever two gpus, while `"2,3"` selects those specific gpus, and `"2,"` selects only the second gpu specifically).

But honestly, setting `CUDA_VISIBLE_DEVICES` is also sufficient given some documentation.

bittremieux:
Superseded by 6ee6351.

bittremieux:
Thanks for the feedback and contributions @gdewael.

PR #286, opened 2024-01-30
Ensure --model has to be set when not training from scratch

gdewael:
I'm adding this pull request directly to the main branch as it does not add extra features.
This PR adds an extra compatibility check between two casanovo options.

Currently, `casanovo train` with a `config.yaml` specifying `train_from_scratch: False`, can still be run without setting the `--model` parameter. (It then still trains from scratch). I've added a `raise ValueError` that makes setting `--model` required when not training from scratch.


bittremieux:
Thanks for the contribution. Incidentally, `train_from_scratch` has actually recently been removed, as it was indeed a superfluous config option. See #275. This will be provided in the next Casanovo release.

Issue #284, opened 2024-01-25
KeyError: '$'


patrick-willems:
Hey,

I am very eager to try out the newest casanovo version on some of our data. While the test MGF file works just fine, I keep running into the same error on our Orbitrap data (after MGF and mzML conversion). Seems like its looking for an amino acid for a '$' character, which doesn't seem to make sense?

```
Predicting DataLoader 0:   2%|▏         | 1/44 [25:28<18:15:27,  0.00it/s]Traceback (most recent call last):
  File "/home/prc/anaconda3/envs/casanovo_env/bin/casanovo", line 8, in <module>
    sys.exit(main())
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/rich_click/rich_command.py", line 126, in main
    rv = self.invoke(ctx)
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/casanovo/casanovo.py", line 142, in sequence
    runner.predict(peak_path, output)
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/casanovo/denovo/model_runner.py", line 163, in predict
    self.trainer.predict(self.model, self.loaders.test_dataloader())
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/lightning/pytorch/trainer/trainer.py", line 864, in predict
    return call._call_and_handle_interrupt(
  File "/home/prc/anaconda3/envs/casanovo_env/lib/python3.10/site-packages/lightning/pytorch/trainer/call.py", line 44, in _call_and_handle_interrupt
    return trainer_fn(*args, **kwargs)
[...]

PR #290, opened 2024-02-09
Stabilize torch.topk() behavior

melihyilmaz:
Addresses #284 

To make sure we get the padding token at index  '0'  as the top scoring token for finished beams at each decoding step, I added a small epsilon of `1e-8` to index '0' in `finished_mask` so that it's not zeroed out like the rest of values on the same row. Again, these zero rows correspond to finished beams and we use masking to avoid extending them with new AA tokens.

This seems to resolve the error with minimal/no overhead (I get the same output on both CPU and GPU for the problematic mgf files mentioned in the issue) but unit tests still need to be added (@bittremieux feel free if you get a chance) and I'm open to moving away from `torch.topk()` if there are suggestions for a more robust or elegant solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Instead of having to specify `train_from_scratch` in the config file, training will proceed from an existing model weights file if this is given as an argument to `casanovo train`.
 
+### Fixed
+
+- Fixed beam search decoding error due to non-deterministic selection of beams with equal scores.
+
 ## [4.0.0] - 2023-12-22
 
 ### Added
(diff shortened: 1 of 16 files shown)


bittremieux (changes requested):
I still don't find it super elegant, but it might be a practical fix for now. I combined some code to remove a few redundant steps, but the fix is the same.

@melihyilmaz Can you try to come up with a unit test that failed before the fix and runs now? So we can ensure that this is properly tested and avoid regression issues in the future.

melihyilmaz:
I add a unit test for `_get_topk_beams()` that only passes with the current fix. I also compared results for a small mgf on CPU vs. GPU and those identical now.

bittremieux (approved):

wsnoble:
You should be able to install from the dev branch like this:

`pip install git+https://github.com/Noble-Lab/casanovo.git@dev`

bittremieux:
@lutfia95 We've now released Casanovo v4.1.0 as well that includes this fix, so you can more conveniently now upgrade from PyPI as well.

PR #300, opened 2024-02-20
Rename max_iters to cosine_schedule_period_iters

### LINKED ISSUE #242: Make the learning rate scheduler more flexible
We should modify the config file so that it contains a group of parameters for controlling the learning rate scheduler. This should give some flexibility to try other schedulers that are provided automatically by pytorch.

Currently, we have this:

```
# Number of warmup iterations for learning rate scheduler
warmup_iters: 100_000
# Max number of iterations for learning rate scheduler
max_iters: 600_000
# Learning rate for weight updates during training
learning_rate: 5e-4
```

The `max_iters` is poorly named, so it should be changed to better reflect what it does. In general, the config file should contain more text describing how the learning rate is controlled.

bittremieux:
Fixes #242.

`cosine_schedule_period_iters` better reflects what this config option does, and has correspondingly been renamed.

The config loader checks whether `max_iters` is specified and automatically remaps it to `cosine_schedule_period_iters`, while warning the user about this renaming.


melihyilmaz (commented):
When I ran the branch locally to fine-tune v4.0 weights, I got the error below:

> optimizer = torch.optim.Adam(self.parameters(), **self.opt_kwargs)
> TypeError: Adam.__init__() got an unexpected keyword argument 'max_iters'

Since `max_iters` is still in the ckpt file, we pick it up as a part of `kwargs`. I think we can manually delete it from the current ckpts but I'm not sure if that'd break anything @bittremieux.

bittremieux:
Ideally we fix it in the code as well, because otherwise other people will have the same problem when they try fine-tuning.

Can you share the full stack trace?

bittremieux:
Ideally we fix it in the code as well. Otherwise other people who try fine-tuning will have the same problem.

I added a unit test to check this and added a fix to remove unrecognized hyperparameters during model loading. This seems to work, but I find it a bit less elegant than the previous fix, because now we're changing config options both in `Config` and in `Spec2Pep`. Do you see any alternative solutions @melihyilmaz @wfondrie?

melihyilmaz (approved):
Everything works for me now locally but I can't think of a more elegant alternative.

PR #301, opened 2024-02-20
Run unit tests on macOS with M1 support

bittremieux:
See [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) for the available (free) runners for GitHub Actions.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest, macos-14]
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


bittremieux:
I tried to set `PYTPYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0`, as suggested by the error message, but that didn't solve the issue. There are (conflicting) reports [here](https://github.com/pytorch/pytorch/issues/105839) that this issue might get solved in PyTorch 2.3, which is [scheduled for release in April](https://dev-discuss.pytorch.org/t/pytorch-release-2-3-0-call-for-features/1872).

As a non-macOS user and lacking familiarity with this error, I currently don't see many other options than to wait until then.

bittremieux:
Apparently [MPS is not actually available](https://github.com/pytorch/pytorch/issues/111449#issuecomment-1767613883) for the GitHub Action runners, so there's no way to test this currently. #327 has disabled running on M1/M2 chips (which will automatically try to use MPS), so at least our tests don't fail anymore.

PR #310, opened 2024-02-29
Add FAQ about MGF index tracking

melihyilmaz:


bittremieux (commented):
This commit is fine.

However, I'm wondering whether instead we should properly document all supported input and output formats (as in #27) on a dedicated page instead? On its own, I don't think this question in the FAQ is that useful.

wsnoble:
I would heartily support that suggestion!

bittremieux:
Superseded by #317.

PR #316, opened 2024-03-25
Allow csv to handle all newlines

mobiusklein:
The `csv` module tries to handle newlines itself. On Windows, this leads to line endings of `\r\r\n` instead of `\r\n`.

Setting `newline=''` produces the intended output on both platforms.


bittremieux (approved):
Thanks! I've changed the base branch to `dev` instead of `main`, because that's where we collect changes while preparing for the next release.

One small comment: could you add an entry to the CHANGELOG as well to record this fix?

mobiusklein:
Thank you. I see that the base branch issue was a bigger mess than I expected. I'll make sure to use the `dev` branch in the future.

bittremieux:
No problem. It's not documented in our contributing guidelines, so there was no way for you to know. We should update the CONTRIBUTING.md to reflect this.

PR #340, opened 2024-06-20
313 save final model

Lilferrit:
In order to insure the final model is always saved the following lines were added to the end of ModelRunner.train:

```python
# Always save final model weights at the end of training
if self.config.model_save_folder_path is not None:
    self.trainer.save_checkpoint(
        os.path.join(
            self.config.model_save_folder_path,
            "train-run-final.ckpt"
        )
    )
```

This implementation was tested using a small training run for the case where val_check_interval is not a factor of the total number of training steps and the case where val_check_interval is greater than the total number of training steps. In both cases the final model checkpoints were saved.


Lilferrit:
Added final epoch number to file name - changed implementation to:

```python
# Always save final model weights at the end of training
if self.config.model_save_folder_path is not None:
    self.trainer.save_checkpoint(
        os.path.join(
            self.config.model_save_folder_path,
            f"train-run-final-{self.trainer.current_epoch}.ckpt"
        )
    )
```

casanovo/denovo/model_runner.py
```diff
@@ -111,6 +111,15 @@ def train(
             self.loaders.val_dataloader(),
         )
 
+        # Always save final model weights at the end of training
+        if self.config.model_save_folder_path is not None:
```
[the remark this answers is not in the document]
bittremieux: **question:** Can `save_checkpoint` correctly handle non-existing paths?

casanovo/denovo/model_runner.py
```diff
@@ -111,6 +111,15 @@ def train(
...
+            self.trainer.save_checkpoint(
+                os.path.join(
```
bittremieux, line 117: **todo:** Prefer to use `pathlib` for consistency with the rest of the code.

casanovo/denovo/model_runner.py
```diff
@@ -111,6 +111,15 @@ def train(
...
+                    self.config.model_save_folder_path,
+                    f"train-run-final-{self.trainer.current_epoch}.ckpt",
```
bittremieux, line 119: **issue:** Use a filename that's consistent with the other checkpoint filenames.

bittremieux (changes requested):
Some minor comments to address.

However, can't we get the same results much easier by setting `enable_checkpointing` always to True when creating the `Trainer` and adding a default `ModelCheckPoint` callback? See the [Trainer documentation](https://pytorch-lightning.readthedocs.io/en/1.5.10/api/pytorch_lightning.trainer.trainer.html). That way we can benefit from letting Lightning handle all of this.

casanovo/denovo/model_runner.py
Lilferrit: In regards to the filename, should the filename be the same format as the validation checkpoints? I.e. should the checkpoints saved at the end of a training epoch also be {epoch}-{step}.cpkt, or should the filename for the train epoch and validation checkpoints be changed to be distinct from one-another?

Lilferrit:
> Some minor comments to address.
> 
> However, can't we get the same results much easier by setting `enable_checkpointing` always to True when creating the `Trainer` and adding a default `ModelCheckPoint` callback? See the [Trainer documentation](https://pytorch-lightning.readthedocs.io/en/1.5.10/api/pytorch_lightning.trainer.trainer.html). That way we can benefit from letting Lightning handle all of this.

Agreed - I've reimplemented this using a `ModelCheckpoint` instead. I will push it upstream once I've done some more testing on my end. However regarding the `enable_checkpointing` operation it looks like this option only adds a default callback if no user defined callbacks are added to `callbacks`, so effectively this would do nothing if the validation `ModelCheckpoint` is added. In order to always save the final model I instead added a new `ModelCheckpoint` that fires at the end of every training epoch.

casanovo/denovo/model_runner.py
bittremieux: I'd say that all checkpoints should have consistent filenames.

bittremieux (changes requested):
Great, I think that this is a better solution.

The remaining thing to do is add unit tests that verify that the final model is saved for different situations (different values of steps and epochs). You could also add some tests to check that the periodic checkpoints are properly created as well.

Lilferrit:
> Great, I think that this is a better solution.
> 
> The remaining thing to do is add unit tests that verify that the final model is saved for different situations (different values of steps and epochs). You could also add some tests to check that the periodic checkpoints are properly created as well.

Sounds good, I added some unit test that ensure the last model weights are saved in the scenarios where val_check_Interval is greater than and not a factor of the number of training steps. Unfortunately since the ModelCheckpoint that saves the model checkpoints at the end of evert epoch deletes the last training epoch's checkpoints (it doesn't touch the validation checkpoints) when a new epoch checkpoint is saved and the CLIRunner.invoke is blocking I couldn't think of a practical way to test whether the periodic checkpoints are saved properly.

tests/conftest.py
```diff
@@ -4,6 +4,7 @@
 import psims
 import pytest
 import yaml
+import math
```
bittremieux, line 7: **todo:** Use the recommend Python import order:
1. Block with standard library imports.
2. Block with external library imports.
3. Block with local application imports.

An then within each block the imports are ordered alphabetically.

So here, the `import math` should be part of a first block of imports, then an empty line, and then these remaining imports.

tests/conftest.py
```diff
@@ -255,8 +255,36 @@ def tiny_config(tmp_path):
...
+@pytest.fixture
+def tiny_config(tmp_path):
+    """A config file for a tiny model."""
+    cfg = _get_default_config(tmp_path)
+    return _write_config_file(cfg, tmp_path)
+
+
+@pytest.fixture
+def tiny_config_interval_greater(tmp_path):
+    """Config file where val_check interval is greater than the number of training steps"""
```
bittremieux, line 277: **todo:** Adhere to maximum line length in docstring (black doesn't do this automatically).

tests/conftest.py
```diff
@@ -255,8 +255,36 @@ def tiny_config(tmp_path):
...
+    cfg = _get_default_config(tmp_path)
+    val_check_interval = 50
+    cfg["val_check_interval"] = val_check_interval
+    return _write_config_file(cfg, tmp_path)
+
+
+@pytest.fixture
+def tiny_config_not_factor(tmp_path):
+    """Config file where val_check interval isn't a factor of the number of training steps"""
```
bittremieux, line 286: **todo:** Adhere to maximum line length in docstring (black doesn't do this automatically).

tests/conftest.py
```diff
@@ -255,8 +255,36 @@ def tiny_config(tmp_path):
...
+    """A config file for a tiny model."""
+    cfg = _get_default_config(tmp_path)
+    return _write_config_file(cfg, tmp_path)
+
+
+@pytest.fixture
+def tiny_config_interval_greater(tmp_path):
+    """Config file where val_check interval is greater than the number of training steps"""
+    cfg = _get_default_config(tmp_path)
+    val_check_interval = 50
```
bittremieux, line 279: **nitpick:** Assigning this to a variable is a bit superfluous imo (here and a few lines down).

tests/conftest.py
```diff
@@ -255,8 +255,36 @@ def tiny_config(tmp_path):
...
+
+@pytest.fixture
+def tiny_config(tmp_path):
+    """A config file for a tiny model."""
+    cfg = _get_default_config(tmp_path)
+    return _write_config_file(cfg, tmp_path)
+
+
+@pytest.fixture
+def tiny_config_interval_greater(tmp_path):
```
bittremieux, line 276: **suggestion:** Rather than creating multiple fixtures that will only be used for a single test (which kind of contradicts the benefit of fixtures), I'd just modify the config for that test specifically. See e.g. [here](https://github.com/Noble-Lab/casanovo/blob/313-save-final-model/tests/unit_tests/test_config.py#L18).

tests/test_integration.py
```diff
@@ -86,6 +92,44 @@ def test_train_and_run(
     assert psms.loc[4, "sequence"] == "PEPTLDEK"
     assert psms.loc[4, "spectra_ref"] == "ms_run[2]:scan=111"
 
+    # Test checkpoint saving when val_check_interval is greater than training steps
```
bittremieux, line 95: **suggestion:** I would split this off in a dedicated test to indicate that you're really only testing the model saving part here.

I also think that this is more of a unit test than an integration test, so I would recommend to execute this as `ModelRunner.train` insted of through the CLI runner..

bittremieux (changes requested):
A few suggestions for the tests.

Lilferrit:
Sounds good, I factored the save final model test into a separate unit test `test_save_final_model` in `test_runner.py`.

bittremieux (approved):

bittremieux:
Great! The final thing to do is update the changelog.

Lilferrit:
Sounds great, I added an entry to the changelog.

PR #341, opened 2024-06-20
mzTab validation fix

Lilferrit:
Changed the charge value that's written to the results mzTab file to an integer value in order to comply with the mzTab standard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+- Precursor charges are now exported as integers instead of floats in the mzTab output file, in compliance with the mzTab specification.
+
 ## [4.2.0] - 2024-05-14
 
 ### Added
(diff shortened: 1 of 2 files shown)


bittremieux (approved):
Looks good. Can you update the changelog as well?

Lilferrit:
> Looks good. Can you update the changelog as well?

Sure thing - I just updated the changelog.

CHANGELOG.md
```diff
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Changed
+
+- The charge value that's written to the results mzTab file is now an integer value to comply
+with the mzTab standard.
```
bittremieux, line 12: **todo:** Remove the newline. Also this should probably go under the "Fixed" rather than "Changed" category.

I'd also rephrase a bit more concrete for end users: "Precursor charges are exported as integers instead of floats in the mzTab output file, according to the mzTab specification."

Lilferrit: Sounds good, I edited the changelog entry.

PR #342, opened 2024-06-20
CLI documentation page generation

Lilferrit:
I implemented automatic documentation generation for the casanovo CLI using the sphinx-click package. This does introduce a new dependency on the sphinx-click python package in order to build the documentation.


docs/cli.rst
```diff
@@ -0,0 +1,17 @@
+Casanovo CLI
+==============
```
bittremieux, line 2: **nitpick:** Align the "=" to the text.

docs/cli.rst
```diff
@@ -0,0 +1,17 @@
...
+
+After installing the Casanovo package and all of its dependencies, the Casanovo CLI
```
bittremieux, line 4: **suggestion:** For nicer git diffs, in text files you can split at the sentence level.

docs/_static/js/cli-page.js
```diff
@@ -0,0 +1,25 @@
+(function () {
```
bittremieux, line 1: **question:** Do we need this Javascript hack? How does it look without/with doing this?

docs/images/configure-help.svg
bittremieux, line 1: **todo:** Merge in the latest `dev` to get proper screenshots.

bittremieux (commented):
Looks fine, just a few small clarifications and the latest changes on `dev` need to be merged in.

docs/_static/js/cli-page.js
Lilferrit: Unfortunately the due to the way that the sphinx-click package is implemented the generated CLI documentation page looks pretty terrible without it.

bittremieux: Can you share a screenshot with and without it? Doesn't it just remove a header?

bittremieux: It's not ideal, but also not entirely terrible imo. And adding Javascript to all pages feels pretty unsatisfying as well.

Apparently there were some efforts to [remove the header with spinx-click](https://github.com/click-contrib/sphinx-click/pull/100), but they have stalled unfortunately.

Can you use [`sphinx-click-process-description`](https://sphinx-click.readthedocs.io/en/latest/usage/#events) to remove the second Casanovo header that comes from the docstring? That should bring us pretty close to the desired layout already.

Lilferrit: I changed the implementation to use an event handler instead, here's what it looks like:

image

docs/conf.py
```diff
@@ -12,9 +12,10 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-# import os
-# import sys
-# sys.path.insert(0, os.path.abspath("."))
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath("."))
```
bittremieux, line 18: **nitpick:** Two blank lines after the imports.

docs/conf.py
```diff
@@ -57,9 +58,13 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ["_static"]
+# html_static_path = ["_static"]
 html_theme_options = {"repository_url": "https//github.com/Noble-Lab/casanovo"}
 
+html_js_files = [
```
bittremieux, line 64: **todo:** This can be removed.

docs/conf.py
```diff
@@ -57,9 +58,13 @@
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ["_static"]
+# html_static_path = ["_static"]
```
bittremieux, line 61: **question:** Does this need to be commented?

docs/sphinx_click_handlers.py
```diff
@@ -0,0 +1,9 @@
+def process_description(app, ctx, lines):
+    if ctx.command.name == "main":
+        del lines[:3]
+
+    print(lines)
```
bittremieux, line 5: **question:** Does this need to be there? The sphinx-click documentation says that the content is modified _in-place_, so I don't think it needs to be printed as well.

bittremieux (changes requested):
Final small comments.

docs/_static/js/cli-page.js
bittremieux: Yes, I think that this implementation is better. And the page doesn't look awful.

docs/conf.py
Lilferrit: AFIK this was originally left as a commented out piece of code in the conf.py template, but I'll go ahead and remove it.

docs/sphinx_click_handlers.py
Lilferrit: My bad, that was some debug code I forgot to remove from the commit.

docs/conf.py
bittremieux: It's currently not commented on `main`, so unless the change is required, I think we can leave it as-is,

bittremieux (approved):

Issue #201, opened 2023-06-28
Add brackets and hyphens to PTMs 


wsnoble:
I think we should modify how Casanovo reports PTMs in its predicted sequences.  Ideally, we would be compatible with the Proforma specification:

https://github.com/HUPO-PSI/ProForma/blob/master/SpecDocument/ProForma_v2_draft15_February2022.pdf

But even if we don't adopt unimod IDs in our spec, we should at least make our formatting compatible with the mass difference approach.  The only change is that each modification mass is enclosed in square brackets, and N-term and C-term mods are followed or preceded by a hyphen.  So for example,  `-17.027C+57.021PSEC+57.021TC+57.021LDTVVR` gets turned into `[-17.027]-C[+57.021]PSEC[+57.021]TC[+57.021]LDTVVR`.

wfondrie:
One of the updates I made in the recent depthcharge update is first-class support for proforma. 

Even mskb-style mods are now handled as proforma internally.

wsnoble:
Does that mean we can close this issue? I.e., will this get handled automatically once the new depthcharge changes are in place, or are additional changes on the casanovo side going to be necessary?

wfondrie:
Yes, but let's leave the issue open and close it with that upgrade.

bittremieux:
Note though that according to the [official mzTab specification](https://github.com/HUPO-PSI/mzTab/blob/master/specification_document-releases/1_0-Proteomics-Release/mzTab_format_specification.pdf), the `sequence` column in the output file should be the unmodified sequence, and modifications should be reported in a `modification` column (page 15–16). This is because mzTab predates ProForma. We currently violate this part of the spec as well because it would be annoying to retokenize the predicted peptide sequences.

wsnoble:
Comet (and Tide) solves this by producing three columns: one with the raw sequence, one with the sequence decorated with mods, and one with just the mods.  I think this might be a nice thing to do in Casanovo.

PR #343, opened 2024-06-21
Enhanced Sequencing Run Logging

Lilferrit:
Implemented additional logging functionality in order to provide an enhanced end of sequence run log. This takes the form of an end of sequencing run report, an example is shown below:

```
INFO: ======= Sequencing Run Report =======
INFO: Sequencing Run Start Timestamp: 1719004519s
INFO: Sequencing Run End Timestamp: 1719004539s
INFO: Time Elapsed: 20s
INFO: Executed Command: D:\anaconda3\envs\casanovo_env\Scripts\casanovo sequence sample_data\sample_preprocessed_spectra.mgf -o foo.mztab
INFO: Executed on Host Machine: DESKTOP-P03U1SR
INFO: Sequencing run date: 06/21/24 14:15:39
INFO: Sequenced 128 spectra
INFO: Sequence Score CMF: {0.0: 96, 0.5: 96, 0.9: 66, 0.95: 60, 0.99: 8}
INFO: Max Sequence Length: 17
INFO: Min Sequence Length: 6
INFO: Max GPU Memory Utilization: 400mb
```

All the data to generate this report is recorded on the fly as inference is conducted (i.e. no file parsing). To facilitate this the functionality of the data submodule was extended to add infastructure for run time logging. This will hopefully make it relatively easy to further extend logging and other io functionality in the future.


bittremieux:
Changed base to `dev`, make sure to reflect that locally.

Lilferrit:
Added more information to the end of run report, namely the number of skipped spectra. Unfortunately the model never sees the spectra that are skipped due to pre-check errors such as having invalid precursor info, so these aren't reflected in the end of run report. They do however appear at other points in the log. A sample log is available below: 

```
INFO: ======= Sequencing Run Report =======
INFO: Sequencing Run Start Timestamp: 1719271695s
INFO: Sequencing Run End Timestamp: 1719271714s
INFO: Time Elapsed: 18s
INFO: Executed Command: D:\anaconda3\envs\casanovo_env\Scripts\casanovo sequence sample_data\sample_preprocessed_spectra.mgf -o data/foo -c casanovo.yaml
INFO: Executed on Host Machine: DESKTOP-P03U1SR
INFO: Sequencing run date: 06/24/24 16:28:34
INFO: Attempted to sequence 128 spectra
INFO: Sequenced 72 spectra
INFO: Skipped 56 spectra
INFO: Sequenced 56.25% of total spectra
INFO: Skipped 43.75% of total spectra
INFO: Score Distribution:
INFO: 54 spectra (75.00%) scored >= 0.0
INFO: 54 spectra (75.00%) scored >= 0.5
INFO: 49 spectra (68.06%) scored >= 0.9
INFO: 46 spectra (63.89%) scored >= 0.95
INFO: 8 spectra (11.11%) scored >= 0.99
INFO: Max Sequence Length: 9
INFO: Min Sequence Length: 6
INFO: Max GPU Memory Utilization: 398mb
```

Unfortunately I don't have access to any sample data that has spectra that are skipped at the inference stage, so I haven't been able to fully test the skipped spectra logging.

bittremieux:
Can you describe a bit what the `logging_io` and `prediction_io` are and how they're intended to be used? You mentioned it during the meeting yesterday, but it would be useful to have it briefly written out as well.

Lilferrit:
> Can you describe a bit what the `logging_io` and `prediction_io` are and how they're intended to be used? You mentioned it during the meeting yesterday, but it would be useful to have it briefly written out as well.

For sure, `prediction_io` contains the definition for the `PredictionWriter` interface as well as the definition for `PredictionMultiWriter`. For the most part, all of the functions in the `PredictionWriter` class were already implemented by `MztabWriter`. The `PredictonWriter` interface defines three member functions, `log_prediction`, `log_skipped_spectra`, and `save` which are all optional to implement by a class that implements the interface. The `PredictionWriter` interface is intended to provide a consistent interface for writing Casanovo sequence predictions to external IO (namely files and loggers).

`PredictionMultiWriter` implements the `PredictionWriter` interface, and maintains an internal list of PredictionWriters. When one of the `PredictionWriter` functions is called on a `PredictionMultiWriter` it'll call that member function on all of the PredictionWriters in it's internal list.

`logger_io` contains the definition of `LogPredictionWriter`, a `PredictionWriter` that is used to generate the end of sequencing run report. It maintains a table of predictions (which contains only the predicted peptide sequence and the search engine score at the time of writing), and writes to this table when `log_prediction` is called. When `save` is called it writes the end of run report to a logger object that is provided via the constructor.

bittremieux:
Imo this is a bit over-engineered. Instead, we can just log directly where specific information is available using the Python loggers, rather than having to pass it all upstream to these new interfaces. Your implementation is precisely with the Python logger does after all: having a universal interface that multiple writers can hook into and write to different outputs. So we don't need to reinvent that.

So I'd just log the PSM statistics in the `MzTabWriter`, which is where you have the PSMs. Skipped spectra are logged in the reader, which is where the skipping actually happens (or will happen after the DepthCharge upgrade). (The current spectrum skipping logging is also incorrect and adds non-negligible overhead, which should be avoid for only logging.) Other stuff can similarly be logged in other relevant locations, etc.

Lilferrit:
Oh ok sounds good - so just to be clear should all of the PSM processing (e.g. the score distribution calculation) and logging happen in the `MzTabWriter`? Also, should skipped spectra logging just be delayed until the depth charge upgrade is implemented?

bittremieux:
Yes, I think that's the logical place, considering that all PSMs are aggregated there.

Of course the over-engineering is my personal feeling. We could ask @wfondrie to weigh in as well.

Logging the number of skipped spectra can indeed be delayed for now.

Lilferrit:
I agree that this solution is overengineered for addressing just the post run logging. My reasoning for implementing it this way is having a PSM io interface would make it relatively easy to extend the PSM io in general. For example if we want to support other output formats like parquet or csv in the future and add a command line option to specify multiple output formats (say you want to write the PSMs to csv and mzTab) it would be relatively easy to do this, or if an end user wants to write PSMs to an external component like a database they could write their own PredictionWriter module to handle this.

bittremieux:
optimization

Lilferrit:
Fair enough, I'll get everything moved to the mzTab writer.

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
 
 import collections
 import csv
+import logging
 import operator
 import os
 import re
+from datetime import datetime
```
bittremieux, line 9: **suggestion:** I prefer `import ...` over `from ... import ...` because it makes the code more self-documenting by explicitly showing where the functions come from when using them. I suggest modifying these imports accordingly.

There can be exceptions of course, here for example for the `typing` module, because e.g. type hints get super verbose otherwise and they're only used in specific locations.

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
 
 import natsort
+import numpy as np
+import torch
+from pandas import DataFrame
 
 from .. import __version__
 from ..config import Config
 
+SCORE_BINS = [0.0, 0.5, 0.9, 0.95, 0.99]
```
bittremieux, line 24: **nitpick:** Two blank lines after the imports.

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+
+logger = logging.getLogger("casanovo")
+
+
+def get_score_bins(
+    results_table: DataFrame, score_bins: List[float]
+) -> Dict[float, int]:
+    """
+    From a list of confidence scores, return a dictionary mapping each confidence score
```
bittremieux, line 33: **todo:** Remember the maximum line width (79) even though black doesn't enforce it directly.

bittremieux, line 33: **todo:** According to the [docstring style guide](https://peps.python.org/pep-0257/), the first line should be a short, concise summary, followed by a blank line, then optionally providing more information in the following lines.

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+    to the number of spectra with a confidence greater than or equal to it.
+
+    Parameters
+    ----------
+        results_table: DataFrame
```
bittremieux, line 38: **nitpick:** Maybe give the score column here directly rather than the full dataframe?

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+        score_bins: List[float]
+            Confidence scores to map
+
+    Returns
+    -------
+        score_bin_dict: Dict[float, int]
+            Dictionary mapping each confidence score to the number of spectra with a confidence
+            greater than or equal to it.
+    """
+    se_scores = results_table["score"].to_numpy()
```
bittremieux, line 49: **suggestion:** [`Series.array`](https://pandas.pydata.org/docs/reference/api/pandas.Series.array.html) is recommended. But I don't think this conversion is even necessary, it can operate on the Series directly?

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+    score_bin_dict = {
+        score: len(se_scores[se_scores >= score]) for score in score_bins
```
bittremieux, line 51: **todo:** You can make this slightly more efficient by skipping the indexing: `(se_scores >= score).sum()`.

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+    }
+    return score_bin_dict
```
bittremieux, line 53: **nitpick:** Ultimately I would basically just rewrite this as a one-liner: `return {score: (scores >= score).sum() for score in bins}`.

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+
+
+def get_peptide_lengths(results_table: DataFrame) -> np.ndarray:
+    """
+    Get a numpy array containing the length of each peptide sequence in results_table
+
+    Parameters
+    ----------
+        results_table: DataFrame
```
bittremieux, line 62: **nitpick:** I would similarly just get the relevant column here instead of the full dataframe.

casanovo/data/ms_io.py
bittremieux, line 38: **todo:** There's an unnecessary indentation here and in the subsequent docstrings for the entries after a heading.

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+
+    Returns
+    -------
+        sequence_lengths: np.ndarray
+            Numpy array containing the length of each sequence, listed in the same order
+            that the sequences are provided in.
+    """
+    # Mass modifications do not contribute to sequence length
+    alpha_re = re.compile("[^a-zA-Z]")
+    filter_fun = lambda x: alpha_re.sub("", x)
```
bittremieux, line 73: **todo:** This can be simplified without the lambda function, the copy, and the `apply` as follows: `return sequences.str.replace(alpha_re, "", regex=True).apply(len)`.

(Assuming that the input is a Series called `sequences`.)

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+            Parsed spectrum match table
+
+    Returns
+    -------
+        sequence_lengths: np.ndarray
+            Numpy array containing the length of each sequence, listed in the same order
+            that the sequences are provided in.
+    """
+    # Mass modifications do not contribute to sequence length
+    alpha_re = re.compile("[^a-zA-Z]")
```
bittremieux, line 72: **todo:** Maybe add a comment that this should be modified if/when PTMs are represented using their names in ProForma notation instead of mass differences. See also [here](https://github.com/Noble-Lab/casanovo/issues/201#issuecomment-1638108183).

casanovo/data/ms_io.py
```diff
@@ -2,17 +2,99 @@
...
+    ----------
+        peptide_lengths: np.ndarray
+            Numpy array containing the length of each sequence
+
+    Returns
+    -------
+        peptide_length_histogram: Dict[int, int]
+            Dictionary mapping each unique peptide length to its frequency
+    """
+    lengths, counts = np.unique(peptide_lengths, return_counts=True)
```
bittremieux, line 95: **todo:** This can be made more efficient with [`np.bincount`](https://numpy.org/doc/stable/reference/generated/numpy.bincount.html) or [`np.histogram`](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html) (for explicit bins). `np.unique` has to sort the list (time complexity `O(n log n)`), whereas the others only needs to do a singly pass through the data (`O(n)`).

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
             self._run_map[filename] = i
 
+    def get_report_dict(self) -> Optional[Dict]:
+        """
+        Generate sequencing run report
+
+        Parameters
+        ----------
+            score_bins: List[float], Optional
+                Confidence scores for creating confidence CMF, see getScoreBins
```
bittremieux, line 240: **nitpick:** Referring to an non-existing function (incorrect name).

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+
+        Returns:
```
bittremieux, line 242: **todo:** Use the correct docstring format.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+            report_gen: Dict
+                Generated report represented as a dictionary, or None if no
+                sequencing predictions were logged
+        """
+        results_table = DataFrame(
```
bittremieux, line 247: **nitpick:** No need really to convert this to a dataframe, you can just use two lists instead.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+        )
+
+        if results_table.empty:
+            return None
+
+        peptide_lengths = get_peptide_lengths(results_table)
+        return {
+            "num_spectra": len(results_table),
+            "score_bins": get_score_bins(results_table, self.score_bins),
+            "max_sequence_length": int(np.max(peptide_lengths)),
```
bittremieux, line 261: **nitpick:** Should be possible to get this for free from the peptide lengths histogram (if the bins are sorted).

bittremieux, line 261: **nitpick:** Normally these should be integers already, so the conversion isn't necessary.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+    def log_run_report(self) -> None:
+        """
+        Log sequencing run report
+        """
+        logger.info("======= Sequencing Run Report =======")
+        if self.start_time is not None:
+            end_time = time()
+            elapsed_time = end_time - self.start_time
+            logger.info(
+                f"Sequencing Run Start Timestamp: {int(self.start_time)}s"
```
bittremieux, line 278: **todo:** The timestamps are not really interpretable. I'd convert this to human-readable strings.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+            )
+            logger.info(f"Sequencing Run End Timestamp: {int(end_time)}s")
+            logger.info(f"Time Elapsed: {int(elapsed_time)}s")
+
+        run_report = self.get_report_dict()
+        run_date_string = datetime.now().strftime("%m/%d/%y %H:%M:%S")
```
bittremieux, line 284: **nitpick:** Use YYYY-MM-DD format to avoid confusing everyone non-American.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+            "score_bins": get_score_bins(results_table, self.score_bins),
+            "max_sequence_length": int(np.max(peptide_lengths)),
+            "min_sequence_length": int(np.min(peptide_lengths)),
+            "median_sequence_length": int(np.median(peptide_lengths)),
+            "peptide_length_histogram": get_peptide_length_histo(
+                peptide_lengths
+            ),
+        }
+
+    def log_run_report(self) -> None:
```
bittremieux, line 269: **suggestion:** Most of this actually doesn't depend on the output writer. Rather the better location seems  `casanovo.py`, where the sequencing is actually executed. And the writer is still available through the model runner to get the PSM stats as well.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+        run_report = self.get_report_dict()
+        run_date_string = datetime.now().strftime("%m/%d/%y %H:%M:%S")
+        logger.info(f"Executed Command: {' '.join(argv)}")
+        logger.info(f"Executed on Host Machine: {gethostname()}")
+        logger.info(f"Sequencing run date: {run_date_string}")
+        num_spectra = 0 if run_report is None else run_report["num_spectra"]
+
+        if run_report is None:
+            logger.warning(
+                f"No predictions were logged, this may be due to an error"
```
bittremieux, line 292: **nitpick:** No f-string needed.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+            )
+        else:
+            logger.info(f"Sequenced {num_spectra} spectra")
+
+        if run_report is not None:
```
bittremieux, line 297: **nitpick:** This can just be part of the previous if-else.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+            logger.info(f"Score Distribution:")
```
bittremieux, line 298: **nitpick:** No f-string needed.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+            for score, pop in sorted(run_report["score_bins"].items()):
+                pop_percentage = 100 * pop / num_spectra
```
bittremieux, line 300: **todo:** You don't need to explicitly convert to a percentage, instead you can use the `%` string formatter. And if the score bins are a NumPy array, you can directly divide the full array by the number of spectra and that gives you the numbers that you need.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+                logger.info(
+                    f"{pop} spectra ({pop_percentage:.2f}%) scored >= {score}"
```
bittremieux, line 302: **nitpick:** Use `≥` instead of `>=`.

bittremieux, line 302: **nitpick:** Format the number of spectra with a thousands separator. Format the score as a float.

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+                )
+
+            logger.info(
```
bittremieux, line 305: **nitpick:** I'd swap the order of min and max sequence length. There's also a bit of an overload between "sequencing" and "sequences", so maybe here better "peptide length"?

casanovo/data/ms_io.py
```diff
@@ -143,10 +230,95 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+                f"Max Sequence Length: {run_report['max_sequence_length']}"
+            )
+            logger.info(
+                f"Min Sequence Length: {run_report['min_sequence_length']}"
+            )
+
+        if torch.cuda.is_available():
+            gpu_util = torch.cuda.max_memory_allocated() / (10**6)
```
bittremieux, line 313: **nitpick:** The `10**6` is not an exact conversion to MB (the correct unit is also with capitalization). Instead you can use bit shifts to do this easily.

bittremieux (changes requested):
A lot of these comments are very small things that you can quickly fix.

One bigger comment is that on reflection the output writer is also not the best place to do this logging. Instead, a better location seems `casanovo.py`, where the sequencing is actually executed. The PSM statistics can still be retrieved from the MztabWritter through the ModelRunner.

Additionally, we could still try to generalize it a bit. The runtime, hostname, GPU memory consumption, etc. are equally relevant to a training run. So maybe some general logging for any Casanovo execution, with then a small adaptation for the sequencing run to include PSM statistics there as well.

casanovo/data/ms_io.py
Lilferrit: Sounds good, I changed the implementation to use bincount instead.

casanovo/data/ms_io.py
Lilferrit: Sg, the socket, sys, and datetime modules are now imported explicitly

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+        PSM predictions
+    start_time : Optional[int], default=None
+        The start time of the sequencing run in seconds since the epoch.
+    end_time : Optional[int], default=None
+        The end time of the sequencing run in seconds since the epoch.
+    score_bins: List[float], Optional
+        Confidence scores for creating confidence score distribution,
+        see get_score_bins
+    """
+    log_run_report(start_time=start_time, end_time=end_time)
```
bittremieux, line 239: **praise:** Nice.

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+                "sequence": [psm[0] for psm in predictions],
+                "score": [psm[2] for psm in predictions],
+            }
+        ),
+        score_bins=score_bins,
+    )
+
+    if run_report is None:
+        logger.warning(
+            f"No predictions were logged, this may be due to an error"
```
bittremieux, line 252: **todo:** Unnecessary f-string.

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+        )
+    else:
+        num_spectra = run_report["num_spectra"]
+        logger.info(f"Sequenced {num_spectra:,} spectra")
+        logger.info("Score Distribution:")
+        for score, pop in sorted(run_report["score_bins"].items()):
+            logger.info(
+                f"{pop:,} spectra ({pop / num_spectra:.2%}) scored >= {score}"
```
bittremieux, line 260: **suggestion:** "≥" instead of ">=".

bittremieux (changes requested):

casanovo/utils.py
```diff
@@ -4,12 +4,20 @@
...
+import socket
+import sys
+import time
+from datetime import datetime
+from typing import Tuple, Dict, List, Optional
 
+import numpy as np
 import psutil
 import torch
+from pandas import DataFrame
```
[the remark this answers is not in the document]
Lilferrit: Changed to `import pandas as pd`

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+    score_bin_dict: Dict[float, int]
+        Dictionary mapping each confidence score to the number of spectra
+        with a confidence greater than or equal to it.
+    """
+    return {
+        score: (results_table["score"] >= score).sum() for score in score_bins
+    }
+
+
+def get_peptide_length_histo(peptide_lengths: np.ndarray) -> Dict[int, int]:
```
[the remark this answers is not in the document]
Lilferrit: Changed to `get_peptide_length_hist`

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+    """
+    Get a dictionary mapping each unique peptide length to its frequency
+
+    Parameters
+    ----------
+        peptide_lengths: np.ndarray
```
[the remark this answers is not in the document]
Lilferrit: Superfluous indentation removed.

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+        return None
+
+    peptide_lengths = get_peptide_lengths(results_table)
+    return {
+        "num_spectra": len(results_table),
+        "score_bins": get_score_bins(results_table, score_bins),
+        "max_sequence_length": int(np.max(peptide_lengths)),
+        "min_sequence_length": int(np.min(peptide_lengths)),
+        "median_sequence_length": int(np.median(peptide_lengths)),
+        "peptide_length_histogram": get_peptide_length_histo(peptide_lengths),
```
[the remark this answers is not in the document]
Lilferrit: In one of the logging issues there was a feature request to have some kind of peptide length histogram in the final log. I introduced this functionality in `log_sequencing_report` by logging the top 5 most frequent peptide lengths.

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+    Returns
+    -------
+    sequence_lengths: np.ndarray
+        Numpy array containing the length of each sequence, listed in the
+        same order that the sequences are provided in.
+    """
+    # Mass modifications do not contribute to sequence length
+    # If PTMs ae represented in ProForma notation this filtering operation
+    # needs to be reimplemented
+    alpha_re = re.compile("[^a-zA-Z]")
```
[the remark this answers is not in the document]
Lilferrit: Good catch, changed to

```
return (
    results_table["sequence"]
    .str.replace(r"[^a-zA-Z]", "", regex=True)
    .apply(len)
)
```

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+    Returns
+    -------
+    report_gen: Dict
+        Generated report represented as a dictionary, or None if no
+        sequencing predictions were logged
+    """
+    if results_table.empty:
+        return None
+
+    peptide_lengths = get_peptide_lengths(results_table)
```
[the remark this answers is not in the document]
Lilferrit: The functions themselves are short, but since the functionality of each of them is fairly unique imo combining them all into one function would make it difficult to thoroughly test each of them. Any thoughts on this would be appreciated.

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+    Parameters
+    ----------
+    start_time : Optional[int], default=None
+        The start time of the sequencing run in seconds since the epoch.
+    end_time : Optional[int], default=None
+        The end time of the sequencing run in seconds since the epoch.
+    """
+    logger.info("======= End of Run Report =======")
+    if (start_time is not None) and (end_time is not None):
+        end_time = time.time()
```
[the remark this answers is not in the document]
Lilferrit: Good catch, changed to `end_time = time.time() if end_time is None else end_time`.

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+        elapsed_time = end_time - start_time
+        start_timestamp = datetime.fromtimestamp(start_time).strftime(
+            "%y/%m/%d %H:%M:%S"
+        )
+        end_timestamp = datetime.fromtimestamp(end_time).strftime(
+            "%y/%m/%d %H:%M:%S"
+        )
+        logger.info(f"Run Start Timestamp: {start_timestamp}")
+        logger.info(f"Run End Timestamp: {end_timestamp}")
+        logger.info(f"Time Elapsed: {int(elapsed_time)}s")
```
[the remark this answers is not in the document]
Lilferrit: Changed the implementation to use timedelta instead, thanks!

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+        sequencing predictions were logged
+    """
+    if results_table.empty:
+        return None
+
+    peptide_lengths = get_peptide_lengths(results_table)
+    return {
+        "num_spectra": len(results_table),
+        "score_bins": get_score_bins(results_table, score_bins),
+        "max_sequence_length": int(np.max(peptide_lengths)),
```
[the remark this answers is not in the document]
Lilferrit: Changed to (thanks!):

```
min_length, med_length, max_length = np.quantile(
    peptide_lengths, [0, 0.5, 1]
)
```

casanovo/utils.py
```diff
@@ -66,3 +74,191 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+            Numpy array containing the length of each sequence
+
+    Returns
+    -------
+        peptide_length_histogram: Dict[int, int]
+            Dictionary mapping each unique peptide length to its frequency
+    """
+    bins = np.bincount(peptide_lengths)
+    return {
+        curr_bin: count for curr_bin, count in enumerate(bins) if count != 0
```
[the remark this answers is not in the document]
Lilferrit: Good call, I changed the implementation to utilize numpy arrays instead of the vanilla python data structures.

casanovo/utils.py
Lilferrit: Removed

casanovo/utils.py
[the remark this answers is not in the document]
bittremieux: Hmm, I'm not sure how useful the 5 most frequent peptide lengths are for the user. Let's discuss this during the call.

casanovo/utils.py
```diff
@@ -66,3 +75,221 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+    peptide_lengths: np.ndarray
+        Numpy array containing the length of each sequence
+
+    Returns
+    -------
+    peptide_length_hist_matrix: np.ndarray
+        A matrix of size (n x 2) where each row vector is
+        (peptide_length, frequency)
+    """
+    frequncies = np.bincount(peptide_lengths).reshape((-1, 1))
```
bittremieux, line 123: **typo:** `frequencies`

casanovo/utils.py
```diff
@@ -66,3 +75,221 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+    Parameters
+    ----------
+    start_time : Optional[int], default=None
+        The start time of the sequencing run in seconds since the epoch.
+    end_time : Optional[int], default=None
+        The end time of the sequencing run in seconds since the epoch.
+    """
+    logger.info("======= End of Run Report =======")
+    if (start_time is not None) and (end_time is not None):
+        end_time = time.time() if end_time is None else end_time
```
bittremieux, line 205: **nitpick:** `end_time` will never be `None` based on the `if` the preceding line.

casanovo/utils.py
```diff
@@ -66,3 +75,221 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+        start_datetime = datetime.fromtimestamp(start_time)
+        end_datetime = datetime.fromtimestamp(end_time)
+        delta_datetime = end_datetime - start_datetime
+        logger.info(
+            "Run Start Timestamp: %s",
```
bittremieux, line 210: **nitpick:** "Time" instead of "Timestamp" (or just omit).

casanovo/utils.py
```diff
@@ -66,3 +75,221 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+        )
+    else:
+        num_spectra = run_report["num_spectra"]
+        logger.info("Sequenced %s spectra", num_spectra)
+        logger.info("Score Distribution:")
+        for score, pop in sorted(run_report["score_bins"].items()):
+            logger.info(
+                "%s spectra (%.2f%%) scored >= %s",
+                pop,
+                pop / num_spectra * 100,
```
bittremieux, line 270: **todo:** I don't think you need to multiple by 100 because you're formatting it as a percentage.

casanovo/utils.py
```diff
@@ -66,3 +75,221 @@ def split_version(version: str) -> Tuple[str, str, str]:
...
+                score,
```
bittremieux, line 271: **nitpick:** You can format the score as a float so that it has a consistent number of digits.

casanovo/utils.py
Lilferrit: I ended up just removing the histogram functionality entirely as discussed in the meeting.

casanovo/utils.py
Lilferrit: From what I can tell it turns out the first percentage sign is escaping the second percentage sign, it doesn't have the same behavior as it does in a f string AFIK.

bittremieux: I didn't know that. 👍

bittremieux (approved):

PR #348, opened 2024-06-27
Fix documentation of --output and updated output settings.

Lilferrit:
In order to introduce some of the features/changes requested in #241 and the comments of PR #276 a branch from the `output` branch was created and some new command line features were introduced.

First, the `--root_ckpt_name` option was added to the train sub-command. This option, if set, will add a root name to the checkpoint files, e.g. if the option is set as `--root_ckpt_name foobar` than the checkpoint filenames will have the format `foobar.epoch=2-step=150000.ckpt`. Otherwise the checkpoint files will have the original default format. The checkpoint files will still be saved to the `model_save_folder_path` set in the config file.

Additionally a `--overwrite_output` boolean flag was added to all of the sub commands which produce .log and .mztab output files.  If this flag isn't set then the Casanovo CLI will raise an error if one of the output files already exists, otherwise it will overwrite the output files with the new output.


Lilferrit:
Duplicate of #372.

Issue #108, opened 2022-12-07
Allow specifying different model weights from GitHub Assets


bittremieux:
Use the `--model` parameter to specify a name of the model weights file from the GitHub Assets.

PR #349, opened 2024-06-28
Download weight file from URL

Lilferrit:
In order to address #108, the `--model` parameter now supports URLs. If the `--model` parameter is a valid URL the Casanovo CLI will first look in the cache for the weight file (cache resolution is done via hashing the URL), and if the weight file isn't cached the CLI will attempt to download the weight file and cache it. The CLI will raise an error if the `--model` parameter is set but it's not a valid URL or file in the local filesystem.

However this implementation does have an issue where if the weight file at a URL is updated, but the CLI has previously downloaded weights from that URL, the new weights won't be downloaded because there will still be a cache hit. To fix this I could introduce a `--force` option that skips the cache and always attempts to download and cache weights given a URL.


casanovo/casanovo.py
```diff
@@ -37,6 +37,8 @@
 import torch
 import tqdm
 from lightning.pytorch import seed_everything
+from hashlib import shake_256
```
bittremieux, line 40: **suggestion:** I prefer to avoid the `from ... import ...` again.

casanovo/casanovo.py
```diff
@@ -37,6 +37,8 @@
...
+from urllib.parse import urlparse
```
bittremieux, line 41: **todo:** Same thing, but this should also be moved to the first block of imports because urllib is a standard library.

casanovo/casanovo.py
```diff
@@ -59,10 +61,9 @@ def __init__(self, *args, **kwargs) -> None:
             click.Option(
                 ("-m", "--model"),
                 help="""
-                The model weights (.ckpt file). If not provided, Casanovo
-                will try to download the latest release.
+                Either the model weights (.ckpt file) or a URL pointing to the model weights
+                file. If not provided, Casanovo will try to download the latest release.
```
bittremieux, line 65: **suggestion:** Maybe add "automatically" add the end here.

casanovo/casanovo.py
```diff
@@ -354,9 +355,10 @@ def setup_model(
     seed_everything(seed=config["random_seed"], workers=True)
 
     # Download model weights if these were not specified (except when training).
+    cache_dir = appdirs.user_cache_dir("casanovo", False, opinion=False)
```
bittremieux, line 358: **suggestion:** Convert this to a `Path` already here and then have all functions that do something with the cache dir have a consistent input argument (instead of `str` vs `Path` currently).

casanovo/casanovo.py
```diff
@@ -371,6 +373,17 @@ def setup_model(
                 "model weights"
             ) from None
 
+    # Download model from URL if model is a valid url
+    is_url = _is_valid_url(model)
+    if (model is not None) and is_url:
```
bittremieux, line 378: **suggestion:** I'd make the flow of the code more explicit here. If the model is retrieved from the cache in the previous lines of code, it won't be `None` anymore. But this code still won't be applicable., so I suggest refactoring the code a bit to avoid confusion.

bittremieux, line 378: **nitpick:** Remove superfluous parentheses.

casanovo/casanovo.py
```diff
@@ -371,6 +373,17 @@ def setup_model(
...
+        model = _get_weights_from_url(model, Path(cache_dir))
+
+    if (model is not None) and (not is_url) and (not Path(model).is_file()):
```
bittremieux, line 381: **nitpick:** Remove superfluous parentheses.

casanovo/casanovo.py
```diff
@@ -371,6 +373,17 @@ def setup_model(
...
+        raise ValueError(
+            f"{model} is not a valid URL or checkpoint file path, "
+            "--model argument must be a URL or checkpoint file path"
```
bittremieux, line 384: **nitpick:** Seems a bit redundant error message,

casanovo/casanovo.py
bittremieux, line 381: **suggestion:** Combine the different conditions:

```
if model is not None:
   if is_url:
        ...
    elif not Path(model).is_file():
        ...
```

casanovo/casanovo.py
```diff
@@ -382,7 +395,76 @@ def setup_model(
     return config, model
 
 
-def _get_model_weights() -> str:
+def _get_weights_from_url(
```
bittremieux, line 398: **nitpick:** I'd move these new functions under `_get_model_weights`, because that one is used first in the code above.

casanovo/casanovo.py
```diff
@@ -382,7 +395,76 @@ def setup_model(
...
+    file_url: Optional[str],
```
bittremieux, line 399: **question:** I don't see a default value for `file_url`, so why does it have an `Optional` type hint?

casanovo/casanovo.py
```diff
@@ -382,7 +395,76 @@ def setup_model(
...
+    cache_dir: Path,
+) -> str:
+    """
+    Attempt to download weight file from URL if weights are not already
```
bittremieux, line 403: **suggestion:** Use proper Python style for docstrings, as mentioned previously.

casanovo/casanovo.py
```diff
@@ -382,7 +395,76 @@ def setup_model(
...
+    cached. Otherwise use cased weights. Downloaded weight files will be
```
bittremieux, line 404: **typo:** "cached"

casanovo/casanovo.py
```diff
@@ -382,7 +395,76 @@ def setup_model(
...
+        model weights cache directory path
+
+    Returns
+    -------
+    str
+        path to cached weights file
+    """
+    os.makedirs(cache_dir, exist_ok=True)
+    url_hash = shake_256(file_url.encode("utf-8")).hexdigest(10)
+    cache_file_name = url_hash + ".ckpt"
```
bittremieux, line 421: **todo:** I'm not a huge fan of this, because this removes the human-interpretable weights file name that will subsequently be used. And for example the information that will be logged, which will make it much more challenging for us to diagnose problems related to specific models.

casanovo/casanovo.py
```diff
@@ -382,7 +395,76 @@ def setup_model(
...
+    cache_file_path = cache_dir / cache_file_name
+
+    if cache_file_path.is_file():
+        logger.info(f"Model weights {file_url} retrieved from local cache")
```
bittremieux, line 425: **todo:** For Python logging, formatting strings with `%` instead of f-strings is preferred. E.g.:

```
logger.info("Downloading model weights file %s from %s", filename, url)
```

This is because f-strings are processed eagerly, whereas logging statements can be executed or not based on the logging level. Using `%` style formatting can thus be (minorly) more efficient.

This might be the case in a few different locations.

casanovo/casanovo.py
```diff
@@ -382,7 +395,76 @@ def setup_model(
...
+        return str(cache_file_path)
+
+    logger.info(f"Model weights {file_url} not in local cache, downloading")
+    file_response = requests.get(file_url)
```
bittremieux, line 429: **todo:** Extract the downloading code from here and `_get_model_weights` and re-use this functionality.

Given that the model weights files can be pretty large and not everyone might have a fast internet connection, the progress bar in the existing code could be quite useful.

casanovo/casanovo.py
```diff
@@ -382,7 +395,76 @@ def setup_model(
...
+        logger.error(
+            f"Server Response: {file_response.status_code}: {file_response.reason}"
+        )
+        raise ConnectionError(f"Failed to download weights file: {file_url}")
+
+    logger.info("Model weights downloaded, writing to cache")
+    with open(cache_file_path, "wb") as cache_file:
+        cache_file.write(file_response.content)
+
+    logger.info("Model weights cached")
```
bittremieux, line 442: **nitpick:** Probably not an info-level log message, It can even be omitted entirely imo.

bittremieux (changes requested):
A few small comments.

Two key things are still missing as well:

1. Documentation of this functionality on readthedocs (maybe on the file formats page).
2. Unit tests for the download behavior.

casanovo/casanovo.py
Lilferrit: When I've worked with hashed objects previously normally for human readable documents like reports, logs, or what have you the object is just referred to the first five characters of the hash. Could something like this work here? Another alternative I've thought of is to allow the user to set a human readable alias for the downloaded file, and the file name would become <url_hash>:<alias>.ckpt.

casanovo/casanovo.py
Lilferrit: My bad, I thought that the Optional type hint was similar to std::optional from C++, meaning in this case that the file_url may or may not be present.

casanovo/casanovo.py
bittremieux: My main concern is not the length of the hash or what it looks like, but rather that any hashing removes the information conveyed by the filename. Given that we have different model weights (tryptic, non-tryptic, trained on different subsets, ...) that are distinguished by their filenames, this is quite important information to be include in the log for debugging purposes. So using the first five characters is not a solution.

Allowing the user to set an alias would address this to some extent, but it's not ideal either. Users don't know what they're doing, and they'll just use random names that we won't know what they mean either. So ideally the original filename should be conserved.

Can this be addressed by keeping the filenames as-is, but saving the hash in a separate file in our cache directory? Then this will get consulted to check whether the file needs to be downloaded again.

bittremieux:
> However this implementation does have an issue where if the weight file at a URL is updated, but the CLI has previously downloaded weights from that URL, the new weights won't be downloaded because there will still be a cache hit.

You could also check the "last-modified" information for the URL and verify whether this is more recent than the download time or not.

casanovo/casanovo.py
bittremieux: True. How about something like this:

- All of the following are files and subdirectories in the Casanovo cache directory.
- An index file that contains a mapping from hashes to file names.
- The weights files are then stored in a subdirectory under their hash. I.e. `[cache_dir]/[hash]/[original_filname.ckpt]`

This conserves the original file names and avoids filename collisions.

casanovo/casanovo.py
Lilferrit: Sounds good, I streamlined the model resolution logic.

Lilferrit:
I adjusted the hashing implementation such that now files downloaded from a user specified URL will now be downloaded to `<cache_dir>/<url_hash>/<file_name>.ckpt` (where filename is the actual filename in the url). Files downloaded automatically by `_get_model_weights` are still downloaded directly to the cache directory, the hash sub-dirs are only created when a custom url is specified.

casanovo/casanovo.py
```diff
@@ -494,5 +501,126 @@ def _get_model_weights() -> str:
...
+            requests.ConnectionError,
+            requests.Timeout,
+            requests.TooManyRedirects,
+        ):
+            logger.warning(
+                "Failed to reach %s to get remote last modified time - using cached file",
+                file_url,
+            )
+
+        if cache_time.st_mtime > url_last_modified:
```
Lilferrit, line 563: **question**: For checking whether the upstream weights file is more recent than the cached weights file do we want to use the file created or the file modified time? AFIK these should always be the same but I'm not sure if there's some edge case that I'm not aware of.

Lilferrit:
> A few small comments.
> 
> Two key things are still missing as well:
> 
> 1. Documentation of this functionality on readthedocs (maybe on the file formats page).
> 2. Unit tests for the download behavior.

In regards to the readthedocs documentation it looks like weight files aren't mentioned at all in the "Input Files" section on the file formats page. Is this something we want to add?

casanovo/casanovo.py
bittremieux: They're not necessarily the same, but for practical intents and purposes I don't think it really matters in this case.

casanovo/casanovo.py
```diff
@@ -494,5 +501,126 @@ def _get_model_weights() -> str:
...
+
+    if cache_file_path.is_file() and not force_download:
+        cache_time = cache_file_path.stat()
+        url_last_modified = 0
+
+        try:
+            file_response = requests.head(file_url)
+            if file_response.ok:
+                if "Last-Modified" in file_response.headers:
+                    url_last_modified = email.utils.parsedate_to_datetime(
```
bittremieux, line 545: **todo:** I'm reluctant to include a library with rather different functionality for just a utility function (even though it's part of the standard libraries). Do we need extra functionality that [`strptime`](https://docs.python.org/3/library/datetime.html#datetime.datetime.strptime) doesn't offer?

casanovo/casanovo.py
```diff
@@ -1,13 +1,17 @@
 """The command line entry point for Casanovo."""
 
 import datetime
+import email.utils
 import functools
+import hashlib
 import logging
+import urllib
```
bittremieux, line 8: **todo:** I don't think this is necessary (rather the relevant import is `urllib.parse` a few lines further down).

casanovo/casanovo.py
```diff
@@ -363,22 +366,32 @@ def setup_model(
...
+                )
+                raise PermissionError(
+                    "GitHub API rate limit exceeded while trying to download the "
+                    "model weights"
+                ) from None
+    else:
+        if _is_valid_url(model):
+            model = _get_weights_from_url(model, cache_dir)
+        elif not Path(model).is_file():
+            raise ValueError(
```
bittremieux, line 391: **todo:** Add a similar logging statement.

docs/file_formats.md
```diff
@@ -11,6 +13,14 @@ When you're ready to use Casanovo for *de novo* peptide sequencing, you can inpu
 All three of the above file formats can be used as input to Casanovo for *de novo* peptide sequencing.
 As the official PSI standard format containing the complete information from a mass spectrometry run, mzML should typically be preferred.
 
+### Model Weights
+
+In addition to MS/MS spectra, Casanova also optionally accepts a model weights (.ckpt) input file when running in training, sequencing or evaluating mode.
```
bittremieux, line 18: **typo:** "Casanovo"

docs/file_formats.md
```diff
@@ -11,6 +13,14 @@ When you're ready to use Casanovo for *de novo* peptide sequencing, you can inpu
...
+If no input weight file is provided, Casanovo will automatically use the most recent compatible weights from the official Casanovo github repo, which will be downloaded and cached if they are not already.
```
bittremieux, line 19: **todo:** We should add a little bit more explanation about the resolving logic for compatible weights (cfr. thread on Slack with Mike Riffle).

tests/unit_tests/test_unit.py
```diff
@@ -108,7 +118,127 @@ def request(self, *args, **kwargs):
         mnk.setattr("appdirs.user_cache_dir", lambda n, a, opinion: tmp_dir)
         mnk.setattr("github.Requester.Requester.requestJsonAndCheck", request)
         with pytest.raises(github.RateLimitExceededException):
-            casanovo._get_model_weights()
+            casanovo._get_model_weights(pathlib.Path(tmp_dir))
+
+
+def test_get_weights_from_url(monkeypatch):
```
bittremieux, line 124: **praise:** Very nice.

casanovo/casanovo.py
```diff
@@ -363,22 +366,32 @@ def setup_model(
...
+                f"{model} is not a valid URL or checkpoint file path, "
```
bittremieux, line 392: **todo:** An invalid URL can still be tested easily.

casanovo/casanovo.py
```diff
@@ -363,22 +366,32 @@ def setup_model(
...
-                "GitHub API rate limit exceeded while trying to download the "
-                "model weights. Please download compatible model weights "
-                "manually from the official Casanovo code website "
-                "(https://github.com/Noble-Lab/casanovo) and specify these "
-                "explicitly using the `--model` parameter when running "
-                "Casanovo."
+    cache_dir = Path(appdirs.user_cache_dir("casanovo", False, opinion=False))
+    if model is None:
+        if not is_train:
+            try:
```
bittremieux, line 372: **question:** Maybe with the new abstractions made as part of this PR, the unit tests for the GitHub download functionality can be revisited? (The tests were disabled because they took a long time and it could lead to the GitHub rate being exceeded, but with similar mocking as in your new tests, this might be avoided?)

Additionally, a test of `setup_model` with all its use cases would be relevant, because this is a non-trivial implementation.

bittremieux (changes requested):
It's starting to look quite nice for a complex piece of code. A few more comments.

casanovo/casanovo.py
Lilferrit: Good call, I changed it to use strptime instead.

casanovo/casanovo.py
Lilferrit: Sounds good, I added a test for `setup_model` and full mocking of the github functionality used in `get_model_weights` in order to be able to run `test_get_model_weights` without actually touching the github api.

bittremieux (approved):
Very nice implementation and great work on the unit tests. 👏

PR #350, opened 2024-07-01
Migration to depthcharge v0.4.8

andradesalazar:
This version of Casanovo is now based on depthcharge v0.4.8 instead of v.0.2.3.


pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
     "appdirs",
     "lightning>=2.0",
     "click",
-    "depthcharge-ms>=0.2.3,<0.3.0",
+    "depthcharge-ms>=0.4.8 ",
```
wfondrie, line 25: **todo:** Let's add a `<0.5.0`.

I don't have breaking changes planned in depthcharge, but its easier to make it more liberal if it becomes unnecessary.

wfondrie (changes requested):
Thank you for this excellent PR 🎉 

As a note for others, PR also adds a few features and changes some functionality. Some that I noticed:
- Checkpoints now have useful file names.
- Early stopping can be enabled.
- The learning rate is logged.
- Model precision can now be changed.
- Various gradient things (clipping and such) can be configured. These are very useful for stability.

I've requested a few, mostly small changes. The biggest thing we need to address now are updates to the unit tests, so that all of our CI checks pass.

@wsnoble, @bittremieux, @melihyilmaz - with as big of a change as this is, you should all take it for a spin and make sure we didn't miss anything!

casanovo/denovo/dataloaders.py
```diff
@@ -18,12 +26,12 @@ class DeNovoDataModule(pl.LightningDataModule):
...
     ----------
-    train_index : Optional[AnnotatedSpectrumIndex]
-        The spectrum index file corresponding to the training data.
-    valid_index : Optional[AnnotatedSpectrumIndex]
-        The spectrum index file corresponding to the validation data.
-    test_index : Optional[AnnotatedSpectrumIndex]
-        The spectrum index file corresponding to the testing data.
+    train_paths : str, optional
+            A spectrum lance path for model training.
+    valid_pathas : str, optional
```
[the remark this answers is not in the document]
andradesalazar: Done.

casanovo/denovo/dataloaders.py
```diff
@@ -64,24 +81,123 @@ def __init__(
...
+        self.custom_field_test_mzml = [
+            CustomField("scans", lambda x: x["id"], pa.string()),
+            CustomField("title", lambda x: x["id"], pa.string()),
+        ]
+        
+        self.custom_field_anno = [CustomField("seq", lambda x: x["params"]["seq"], pa.string())]
+
+    def make_dataset(self, paths, annotated, mode, shuffle):
+        """
+        Make spectrum datasets
```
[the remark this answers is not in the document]
andradesalazar: Done

casanovo/denovo/dataloaders.py
```diff
@@ -146,68 +278,37 @@ def _make_loader(
...
-    precursor_mzs = torch.tensor(precursor_mzs)
-    precursor_charges = torch.tensor(precursor_charges)
-    precursor_masses = (precursor_mzs - 1.007276) * precursor_charges
-    precursors = torch.vstack(
-        [precursor_masses, precursor_charges, precursor_mzs]
-    ).T.float()
-    return spectra, precursors, np.asarray(spectrum_ids)
+    spectrum._inner._intensity = spectrum.intensity / np.linalg.norm(
+        spectrum.intensity
+    )
```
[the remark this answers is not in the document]
andradesalazar: Yes, we were also thinking the parentheses are misplaced in Depthcharge. If you change that version, I would directly import the function from Depthcharge and leave it as it is otherwise.

pyproject.toml
andradesalazar: Makes sense. Done!

casanovo/denovo/dataloaders.py
```diff
@@ -64,24 +81,123 @@ def __init__(
...
+                    path=lance_path,
+                    parse_kwargs=parse_kwargs,
+                    **dataset_params,
+                )
+    
+        if shuffle:
+            dataset = ShufflerIterDataPipe(
+                dataset,
+                buffer_size=self.buffer_size
+            )
```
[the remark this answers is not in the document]
andradesalazar: I also found it very cool! With the current version of the training dataset in the mgf file the spectra are somehow sorted by peptide. That's why one for sure needs to set it up, as the training does not converge otherwise.

andradesalazar:
Hi @wfondrie ,

thanks for the comments :) 

are you taking care of the updates to the unit tests, so that the CI checks pass or is it better if I have a look? 

I think the documentation probably needs to be updated a bit, as well as the download of the latest weights for prediction, as the old ones are not compatible anymore. 

Best,
Daniela

bittremieux:
> I think the documentation probably needs to be updated a bit, as well as the download of the latest weights for prediction, as the old ones are not compatible anymore.

Yes, we'll cut a new release v5.x.x for this implementation, as these are some breaking changes. We'll have to train a new model, but with the new major version the downloading code won't get confused.

> are you taking care of the updates to the unit tests, so that the CI checks pass or is it better if I have a look?

Some of the first fixes might be relatively straightforward, with some renamed modules that have to be updated in the unit tests. If you have some bandwidth to look at it, feel free to do so.

casanovo/denovo/model_runner.py
```diff
@@ -107,6 +114,24 @@ def __init__(
             ),
         ]
 
+        # Configure early stopping
+        if config.early_stopping_patience is not None:
+            self.callbacks.append(
+                EarlyStopping(
```
Lilferrit, line 120: Removing for now. This will be introduced back into `casanovo/dev` by a future pr if it is decided to introduce early stopping functionality into the mainline casanovo release.

casanovo/denovo/model_runner.py
```diff
@@ -107,6 +114,24 @@ def __init__(
...
+                    patience=self.config.early_stopping_patience,
+                    verbose=True,
+                    check_finite=True,
+                    mode="min",
+                )
+            )
+        # Configure learning rate monitor
+        if config.tb_summarywriter is not None:
+            self.callbacks.append(
+                LearningRateMonitor(logging_interval="step", log_momentum=True)
```
Lilferrit, line 132: Same thing here, removing for now as this will be reintroduced in an open pr.

tests/unit_tests/test_unit.py
```diff
@@ -418,18 +416,6 @@ def test_is_valid_url():
     assert not casanovo._is_valid_url("foobar")
 
 
-def test_tensorboard():
```
Lilferrit, line 421: Since this branch uses PyLighting's implementation of tensorboard, I have removed this test.

casanovo/data/ms_io.py
```diff
@@ -184,8 +184,11 @@ def save(self) -> None:
                 ),
                 1,
             ):
-                filename = os.path.abspath(psm.spectrum_id[0])
+                filename = psm.spectrum_id[0]
                 idx = psm.spectrum_id[1]
+                if Path(filename).suffix == ".mgf" and idx.isnumeric():
+                    idx = f"index={idx}"
```
Lilferrit, line 190: This is a temporary workaround I added to get the integration test to run, since it doesn't look like the `scan_id` field of a `SpectrumDataset` doesn't have the `index=` prefix when a spectrum comes from an mgf file (it has the `scan=` prefix for mzml files though). It seems like this is a DepthCharge issue so a cleaner fix for this would probably have to happen in DepthCharge.

tests/unit_tests/test_unit.py
```diff
@@ -938,12 +940,8 @@ def test_run_map(mgf_small):
...
     # Set peak file by base file name only.
     out_writer.set_ms_run([os.path.basename(mgf_small.name)])
-    assert os.path.basename(mgf_small.name) not in out_writer._run_map
-    assert os.path.abspath(mgf_small.name) in out_writer._run_map
-    # Set peak file by full path.
-    out_writer.set_ms_run([os.path.abspath(mgf_small.name)])
-    assert os.path.basename(mgf_small.name) not in out_writer._run_map
-    assert os.path.abspath(mgf_small.name) in out_writer._run_map
+    assert mgf_small.name in out_writer._run_map
+    assert os.path.abspath(mgf_small.name) not in out_writer._run_map
```
Lilferrit, line 944: I updated this test to reflect the current behavior of `MztabWriter`, but it might be worth looking into we want to change the behavior of `MztabWriter`, especially if depthcharge is updated to include the full path in the spectrum dataloaders.

Lilferrit:
It looks like there's a bug in `Spec2Pep._finish_beams` where beams that have not been predicted to end aren't checked for early termination due to exceeding the precursor m/z tolerance if the tokenizer doesn't have any residues with negative mass.

Particularly, this loop

```python
for aa in ([None] if finished_beams[i] else aa_neg_mass_idx):
```

that does the early termination check is never entered if the beam isn't finished and there is nothing in `aa_neg_mass_idx`.

Lilferrit:
I've initialized the model's tokenizer with the residues from the tiny config as a work around to get the `test_beam_search_decode` test to run as we look into potential fixes, assuming this actually is a bug.

bittremieux:
Yes, from a quick check I think you're right.

[In the current version](https://github.com/Noble-Lab/casanovo/blob/main/casanovo/denovo/model.py#L349), there's always at least `None` included for no AAs with a negative mass. This is missing in the changed version.

tests/unit_tests/test_unit.py
```diff
@@ -701,26 +723,34 @@ def test_beam_search_decode():
...
-    model = Spec2Pep(n_beams=3, residues="massivekb", min_peptide_len=3)
+    model = Spec2Pep(
+        n_beams=3,
+        min_peptide_len=3,
+        tokenizer=depthcharge.tokenizers.peptides.PeptideTokenizer(
+            residues=config.residues
+        ),
+    )
     beam = model.n_beams  # S
-    model.decoder.reverse = True
```
Lilferrit, line 706: It doesn't look like the current `PeptideDecoder` supports the reverse option.

bittremieux:
The tests still seem to fail on GitHub, in contrast to the latest comment message. @Lilferrit is this expected behavior?

Lilferrit:
Best I can tell looking at the GitHub actions logs, the reason the tests fail on GitHub is due to the Pylance `hf_converter` keyword issue in Depthcharge. The test do pass in my local environment, but I manually downgraded Pylance to v0.15.0. Should I update Casanovo's `pyproject.toml` to require Pylance v0.15.0? That should solve the issue with the tests not passing on GitHub.

bittremieux:
Yes.

Lilferrit:
> Yes. And also make an issue and link it to the one in DepthCharge to track this.

Done, all of the test on GitHub pass now, and seem to run faster than they have historically as well.

Lilferrit:
I overwrote the previous merge with dev with the latest rebase to make the diff more representative of any new functionality.

Issue #356, opened 2024-07-10
Greater than or equal to character ≥ causes logging error in end of run report.


Lilferrit:
The greater than or equal to sign seems to be causing a logging error in the end of run report. I'm running Casanovo on Windows so this may be a platform specific bug. Regardless, here is the end of run report logging output:

```
INFO: ======= End of Run Report =======
INFO: Run Start Time: 24/07/10 16:32:03
INFO: Run End Time: 24/07/10 16:32:23
INFO: Time Elapsed: 0:00:19.993346
INFO: Executed Command: D:\anaconda3\envs\casanovo_env\Scripts\casanovo sequence sample_data\sample_preprocessed_spectra.mgf -o foo
INFO: Executed on Host Machine: DESKTOP-P03U1SR
INFO: Max GPU Memory Utilization: 381 MiB
INFO: Sequenced 128 spectra
INFO: Score Distribution:
INFO: 96 spectra (75.00%) scored ≥ 0.00
--- Logging error ---
Traceback (most recent call last):
  File "D:\anaconda3\envs\casanovo_env\lib\logging\__init__.py", line 1103, in emit
    stream.write(msg + self.terminator)
  File "D:\anaconda3\envs\casanovo_env\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u2265' in position 109: character maps to <undefined>
Call stack:
  File "D:\anaconda3\envs\casanovo_env\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "D:\anaconda3\envs\casanovo_env\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\anaconda3\envs\casanovo_env\Scripts\casanovo.exe\__main__.py", line 7, in <module>
    sys.exit(main())
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\rich_click\rich_command.py", line 367, in __call__
    return super().__call__(*args, **kwargs)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\rich_click\rich_command.py", line 152, in main
    rv = self.invoke(ctx)
[...]

PR #357, opened 2024-07-10
End of run report logging error fix

Lilferrit:
Logging statement fix for #356


Lilferrit:
I just tried the new fix on my local Windows environment and the bug seems to be resolved. I'm glad it worked because the GEQ does look cleaner IMO - I'll go ahead and merge.

Issue #207, opened 2023-07-10
Warn for annotated spectra with unknown sequence formatting


wsnoble:
Anyone got a guess as to what this error might mean?  I am training on an MGF containing 2,373,619 spectra with Casanovo v3.4.0. According to the stderr output, the training read in 2,373,619 spectra and then died with this traceback:

```
Traceback (most recent call last):
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/bin/casanovo", line 8, in <module>
    sys.exit(main())
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/casanovo/casanovo.py", line 173, in main
    model_runner.train(peak_path, peak_path_val, model, config)
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/casanovo/denovo/model_runner.py", line 229, in train
    train_index = AnnotatedSpectrumIndex(
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/depthcharge/data/hdf5.py", line 445, in __init__
    super().__init__(
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/depthcharge/data/hdf5.py", line 104, in __init__
    self.add_file(ms_file)
  File "/net/noble/vol1/home/noble/miniconda3/envs/casanovo/lib/python3.10/site-packages/depthcharge/data/hdf5.py", line 230, in add_file
    group.create_dataset(
[...]

PR #359, opened 2024-07-31
Eliminate evaluate Command

Lilferrit:
Eliminated the `evaluate` command in favor of a `--evaluate` command line option for the `sequence` command. Evaluation metrics will still be logged to the console as before if the `--evaluate` options is set. The model (`Spec2Pep`) will also log predictions to its out_writer in validation mode similar as it does in prediction mode.


bittremieux (changes requested):
This works, but I find the changes in `model.py` a bit messy and unsatisfying.

If the goal of the evaluation is the get the peptide and amino acid metrics, can this not be simplified by:
- First just do standard predictions.
- After the whole inference has been finished, you then have all of the peptides for each spectrum. Then you can read the peptide sequences from the annotated MGF separately, and compare these to each other.
- This removes all of the `validation` related complexity from the model and should simplify the flow of the data considerably.
- The evaluation part is also more maintainable, and it would for example be much easier to add another data source for evaluation (e.g. another mzTab file with "ground truth" rather than an annotated MGF).

Note that it's slightly different from the current validation approach, because that also gives the loss, which you wouldn't have in this approach. However, it seems to me that the loss is not that informative anyway, and not something a user would expect to get when specifying `evaluation` during prediction.

The best approach to tackle this should probably be discussed.

Lilferrit:
> This works, but I find the changes in `model.py` a bit messy and unsatisfying.
> 
> If the goal of the evaluation is to get the peptide and amino acid metrics, can this not be simplified by:
> 
> * First just do standard predictions.
> * After the whole inference has been finished, you then have all of the peptides for each spectrum. Then (at a higher level than inside of the model) you can read the peptide sequences from the annotated MGF separately, and compare these to each other.
> * This removes all of the `validation` related complexity from the model and should simplify the flow of the data considerably.
> * The evaluation part is also more maintainable, and it would for example be much easier to add another data source for evaluation (e.g. an mzTab or CSV file with "ground truth" rather than an annotated MGF).
> 
> Note that it's slightly different from the current validation approach, because that also gives the loss, which you wouldn't have in this approach. However, it seems to me that the loss is not that informative anyway, and not something a user would expect to get when specifying `evaluate` during prediction.
> 
> The best approach to tackle this should probably be discussed.

I agree that this approach makes more sense. Tentatively what I'm thinking is that `ModelRunner` should handle the validation metrics calculation using the model output and the spectra annotations, using the functionality in `evaluate.py`. This would probably leave the actual model itself unchanged.

casanovo/denovo/model_runner.py
```diff
@@ -116,36 +116,25 @@ def train(
...
-        self.initialize_model(train=False)
-
-        test_index = self._get_index(peak_path, True, "evaluation")
-        self.initialize_data_module(test_index=test_index)
-        self.loaders.setup(stage="test", annotated=True)
-
-        self.trainer.validate(self.model, self.loaders.test_dataloader())
-
-    def predict(self, peak_path: Iterable[str], output: str) -> None:
+    def predict(
```
Lilferrit, line 119: **question**: Is it alright to have all of the model output and spectrum annotations in memory at once or does this need to be done in batches?

Lilferrit:
I reimplemented evaluate mode by having `ModelRunner` handle the evaluation instead of using the validation functionality of the `Spec2Pep` model itself.

casanovo/denovo/model_runner.py
```diff
@@ -116,36 +118,52 @@ def train(
...
-        self.initialize_trainer(train=False)
-        self.initialize_model(train=False)
-
-        test_index = self._get_index(peak_path, True, "evaluation")
-        self.initialize_data_module(test_index=test_index)
-        self.loaders.setup(stage="test", annotated=True)
+        model_output = [psm[0] for psm in self.writer.psms]
+        spectrum_annotations = [
+            test_index[i][4] for i in range(test_index.n_spectra)
+        ]
```
Lilferrit, line 136: **question**: Is it alright to have all of the spectrum annotations and predictions in memory at once or does this need to be done in batches?

casanovo/casanovo.py
```diff
@@ -128,64 +128,50 @@ def main() -> None:
...
     output: Optional[str],
     verbosity: str,
+    evaluate: bool,
 ) -> None:
     """De novo sequence peptides from tandem mass spectra.
 
-    PEAK_PATH must be one or more mzMl, mzXML, or MGF files from which
-    to sequence peptides.
+    PEAK_PATH must be one or more mzML, mzXML, or MGF files from which
+    to sequence peptides. If evaluate is set to true peak_path must be
```
bittremieux, line 154: **nitpick:** `True` and PEAK_PATH.

casanovo/denovo/model_runner.py
```diff
@@ -116,36 +118,52 @@ def train(
             self.loaders.val_dataloader(),
         )
 
-    def evaluate(self, peak_path: Iterable[str]) -> None:
-        """Evaluate peptide sequence preditions from a trained Casanovo model.
+    def log_metrics(self, test_index: AnnotatedSpectrumIndex) -> None:
+        """Log pep_precision and aa_precision
```
bittremieux, line 122: **nitpick:** Write out the actual words instead of variable names, which are slightly less informative.

casanovo/denovo/model_runner.py
bittremieux: That's fine. The size of the PSMs is going to be much less than the actual model, so no need to over-optimize this small part.

casanovo/denovo/model_runner.py
```diff
@@ -116,36 +118,52 @@ def train(
...
+        aa_precision, _, pep_precision = aa_match_metrics(
+            *aa_match_batch(
+                spectrum_annotations,
+                model_output,
+                depthcharge.masses.PeptideMass().masses,
+            )
+        )
 
-        self.trainer.validate(self.model, self.loaders.test_dataloader())
+        logger.info("Peptide Precision: %f", pep_precision)
```
bittremieux, line 145: **nitpick:** Round to a fixed number of decimals for pretty printing. Potentially format as a percentage instead of a float.

bittremieux (changes requested):
> My only concern at this point is how to go about testing the evaluation metric calculations. The simplest solution I can think of is to have log_metrics also return the aa_precision and pep_precision in addition to the logging operation. From there I could introduce some unit tests.

I don't think that we need to check the actual metric values. We have dedicated unit tests that already do that.

What should be added is some tests that verify correct behavior without/with evaluation specified based on different types of input files (annotated vs simple MGF, mzML).

casanovo/denovo/model_runner.py
Lilferrit: Unfortunately it doesn't look like logging statements support percentages natively (I ran into the same thing when implementing the end of run report logging) so I just manually multiply the values by 100 instead.

Lilferrit:
> What should be added is some tests that verify correct behavior without/with evaluation specified based on different types of input files (annotated vs simple MGF, mzML).

Sounds good, I'll look into getting some tests implemented for these cases.

Lilferrit:
I did some experimenting while trying to set up the test cases, and it looks like in the current implementation of `ModelRunner._get_index` when the argument `annotated` is set to true unannotated peak files (for the most part) are silently ignored. From a quick glance this ultimately stems from the behavior of `_get_peak_filenames`. However, in the case of unannotated mgf files it looks like there is an uncaught exception from the Depth Charge library that gets kicked up to the `ModelRunner` when attempting to construct an `AnnotatedSpectrumIndex`. I've added the error itself to the end of this comment.

Imo silently ignoring unannotated files is not desirable behavior in the case of running model evaluation post sequencing (this also means that the unannotated files would simply not get sequenced). The best way that comes to mind of getting around this issue is to check that all of the peak files are annotated before sequencing begins, and throwing an exception if any of them aren't. However I'm not sure if there's a quick and easy way to do this.

I put the test cases in progress on the branch `elim-eval-tests` if anyone wants to take a look.

```
casanovo\denovo\model_runner.py:183: in predict
    test_index = self._get_index(peak_path, evaluate, "")
casanovo\denovo\model_runner.py:420: in _get_index
    return Index(index_fname, filenames, valid_charge=valid_charge)
D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py:446: in __init__
    super().__init__(
D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py:104: in __init__
    self.add_file(ms_file)
D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py:231: in add_file
    group.create_dataset(
D:\anaconda3\envs\casanovo_env\lib\site-packages\h5py\_hl\group.py:183: in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
D:\anaconda3\envs\casanovo_env\lib\site-packages\h5py\_hl\dataset.py:166: in make_new_dset
    dset_id.write(h5s.ALL, h5s.ALL, data)
h5py\\_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py\\_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
h5py\\h5d.pyx:282: in h5py.h5d.DatasetID.write
    ???
h5py\\_proxy.pyx:147: in h5py._proxy.dset_rw
    ???
h5py\\_conv.pyx:442: in h5py._conv.str2vlen
    ???
h5py\\_conv.pyx:96: in h5py._conv.generic_converter
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   ???
E   TypeError: Can't implicitly convert non-string objects to strings

h5py\\_conv.pyx:247: TypeError
```

casanovo/denovo/model_runner.py
bittremieux: Oh right, I already forgot about the lack of percentage support. Consistent decimal formatting is still good though. 👍

bittremieux:
> Imo silently ignoring unannotated files is not desirable behavior in the case of running model evaluation post sequencing (this also means that the unannotated files would simply not get sequenced).

I agree.

> The best way that comes to mind of getting around this issue is to check that all of the peak files are annotated before sequencing begins, and throwing an exception if any of them aren't. However I'm not sure if there's a quick and easy way to do this.

That's tricky, because there's indeed no elegant way to do this, so I'm not really in favor of trying to hack this in.

Instead, giving better error messages is a good starting point. Then at least users will know what the problem is and how to fix it if they want to run evaluation.

casanovo/denovo/model_runner.py
```diff
@@ -394,9 +415,15 @@ def _get_index(
         else:
             index_fname = Path(self.tmp_dir.name) / f"{uuid.uuid4().hex}.hdf5"
 
-        Index = AnnotatedSpectrumIndex if annotated else SpectrumIndex
-        valid_charge = np.arange(1, self.config.max_charge + 1)
-        return Index(index_fname, filenames, valid_charge=valid_charge)
+        try:
```
Lilferrit, line 418: I'm not super happy with this, but unfortunately depth charge raises a `ValueError` exception when the `AnnotatedSpectrumIndex` constructor is called with an unnonotated MGF file is in the peak file list. I figured the best thing to do would to raise a more helpfull exception, but I'm definately open to a better solution.

tests/unit_tests/test_runner.py
```diff
@@ -174,3 +175,59 @@ def test_save_final_model(tmp_path, mgf_small, tiny_config):
...
+def test_evaluate(
+    tmp_path, mgf_small, mzml_small, mgf_small_unannotated, tiny_config
+):
+    """Test model evaluation during sequencing"""
+    # Train tiny model
+    config = Config(tiny_config)
+    config.max_epochs = 1
+    model_file = tmp_path / "epoch=0-step=1.ckpt"
+    with ModelRunner(config) as runner:
+        runner.train([mgf_small], [mgf_small])
```
bittremieux, line 189: **question:** I don't think training is really necessary for this test, we can just use a randomly initialized model, if that's compatible with our `ModelRunner` interface.

casanovo/denovo/model_runner.py
bittremieux: The difference between the `FileNotFoundError`, `ValueError`, and `RuntimeWarning` as checked in the tests also seems a bit weird, because these are essentially the same situations. Can this be made more consistent?

Optionally we can request/implement changes in DepthCharge as well.

Lilferrit: Another option is that it should be relatively easy to check whether a MGF file is annotated or not, and filter it out if it isn't annotated. A quick implementation I can think of is to do a pass through the file and make sure that the number of annotations is equal to the number of spectra (i.e. the number of lines that begin with "BEGIN IONS" is equal to the number of lines that begin with "SEQ="). This would eliminate the `ValueError` condition, leaving just the `FileNotFoundError` and `RuntimeWarning`. The `FileNotFound` error would still be thrown if no peak files were found,  under the given peak file path(s), and `RuntimeWarning` would be thrown whenever a file is filtered.

I wish there was a more specific class of warning than `RuntimeWarning` to use in this case. I guess we could implement our own warning class, but I'm not sure that would be worth it just for this use case.

tests/unit_tests/test_runner.py
Lilferrit: Training for one epoch was a workaround in order to get randomly initialized model weights. AFIK it's impossible to initialize the `ModelRunner` with random weights except for during training; in any other case model weights have to be provided.

casanovo/denovo/model_runner.py
Lilferrit: I suppose another option could be to leave this:

```python
Index = AnnotatedSpectrumIndex if annotated else SpectrumIndex
valid_charge = np.arange(1, self.config.max_charge + 1)
return Index(index_fname, filenames, valid_charge=valid_charge)
```

as is and simply let depth charge throw the `TypeError` exception without catching it. Considering that #207 seems to be a very similar error related to missing `SEQ=` in an annotated MGF file, and was marked as a depth charge v0.4 milestone, imo this isn't a bad way to go - basically treating this specific issue as a depth charge issue.

As for keeping the try/except block and checking to make sure that the `TypeError` is related to trying to construct an `AnnotatedSpectrumIndex` using a unannotated MGF file, I am not quite sure on a good way of approaching this.

bittremieux: > As for keeping the try/except block and checking to make sure that the TypeError is related to trying to construct an AnnotatedSpectrumIndex using a unannotated MGF file, I am not quite sure on a good way of approaching this.

You'd need to check the error message, although it's not the most robust approach.

We can discuss desired behavior during the call.

Lilferrit: I tried running evaluate with a unannotated MGF file using the current dev version of Casanovo, and I got the same error:

```
(casanovo_env) C:\Users\gwen\OneDrive\Documents\casanovo>casanovo evaluate data\sample_unannotated.mgf
WARNING: Dataloader multiprocessing is currently not supported on Windows or MacOS; using only a single thread.
Seed set to 454
INFO: Model weights file C:\Users\gavin\AppData\Local\casanovo\casanovo_v420_v4_2_0.ckpt retrieved from local cache
INFO: Casanovo version 4.1.1.dev15+gcedfaa7.d20240625
INFO: Sequencing and evaluating peptides from:
INFO:   data\sample_unannotated.mgf
GPU available: True (cuda), used: True
TPU available: False, using: 0 TPU cores
HPU available: False, using: 0 HPUs
INFO: Reading 1 files...
data\sample_unannotated.mgf: 128spectra [00:00, 2212.82spectra/s]
Traceback (most recent call last):
  File "D:\anaconda3\envs\casanovo_env\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "D:\anaconda3\envs\casanovo_env\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\anaconda3\envs\casanovo_env\Scripts\casanovo.exe\__main__.py", line 7, in <module>
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\rich_click\rich_command.py", line 367, in __call__
    return super().__call__(*args, **kwargs)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\rich_click\rich_command.py", line 152, in main
    rv = self.invoke(ctx)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "C:\Users\gavin\OneDrive\Documents\casanovo\casanovo\casanovo.py", line 185, in evaluate
    runner.evaluate(annotated_peak_path)
  File "C:\Users\gavin\OneDrive\Documents\casanovo\casanovo\denovo\model_runner.py", line 136, in evaluate
    test_index = self._get_index(peak_path, True, "evaluation")
  File "C:\Users\gavin\OneDrive\Documents\casanovo\casanovo\denovo\model_runner.py", line 401, in _get_index
    return Index(index_fname, filenames, valid_charge=valid_charge)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py", line 446, in __init__
    super().__init__(
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py", line 104, in __init__
    self.add_file(ms_file)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py", line 231, in add_file
    group.create_dataset(
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\h5py\_hl\group.py", line 183, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\h5py\_hl\dataset.py", line 166, in make_new_dset
    dset_id.write(h5s.ALL, h5s.ALL, data)
  File "h5py\\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\\h5d.pyx", line 282, in h5py.h5d.DatasetID.write
  File "h5py\\_proxy.pyx", line 147, in h5py._proxy.dset_rw
  File "h5py\\_conv.pyx", line 442, in h5py._conv.str2vlen
  File "h5py\\_conv.pyx", line 96, in h5py._conv.generic_converter
  File "h5py\\_conv.pyx", line 247, in h5py._conv.conv_str2vlen
TypeError: Can't implicitly convert non-string objects to strings
```

The same thing happens with the train command as well:

```
(casanovo_env) C:\Users\gwen\OneDrive\Documents\casanovo>casanovo train data\sample_unannotated.mgf -p data\sample_unannotated.mgf 
WARNING: Dataloader multiprocessing is currently not supported on Windows or MacOS; using only a single thread.
Seed set to 454
INFO: Casanovo version 4.1.1.dev15+gcedfaa7.d20240625
INFO: Training a model from:
INFO:   data\sample_unannotated.mgf
INFO: Using the following validation files:
INFO:   data\sample_unannotated.mgf
GPU available: True (cuda), used: True
TPU available: False, using: 0 TPU cores
HPU available: False, using: 0 HPUs
INFO: Reading 1 files...
data\sample_unannotated.mgf: 128spectra [00:00, 3714.21spectra/s]
Traceback (most recent call last):
  File "D:\anaconda3\envs\casanovo_env\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "D:\anaconda3\envs\casanovo_env\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "D:\anaconda3\envs\casanovo_env\Scripts\casanovo.exe\__main__.py", line 7, in <module>
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\rich_click\rich_command.py", line 367, in __call__
    return super().__call__(*args, **kwargs)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\rich_click\rich_command.py", line 152, in main
    rv = self.invoke(ctx)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\click\core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "C:\Users\gavin\OneDrive\Documents\casanovo\casanovo\casanovo.py", line 232, in train
    runner.train(train_peak_path, validation_peak_path)
  File "C:\Users\gavin\OneDrive\Documents\casanovo\casanovo\denovo\model_runner.py", line 110, in train
    train_index = self._get_index(train_peak_path, True, "training")
  File "C:\Users\gavin\OneDrive\Documents\casanovo\casanovo\denovo\model_runner.py", line 401, in _get_index
    return Index(index_fname, filenames, valid_charge=valid_charge)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py", line 446, in __init__
    super().__init__(
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py", line 104, in __init__
    self.add_file(ms_file)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\depthcharge\data\hdf5.py", line 231, in add_file
    group.create_dataset(
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\h5py\_hl\group.py", line 183, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
  File "D:\anaconda3\envs\casanovo_env\lib\site-packages\h5py\_hl\dataset.py", line 166, in make_new_dset
    dset_id.write(h5s.ALL, h5s.ALL, data)
  File "h5py\\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\\h5d.pyx", line 282, in h5py.h5d.DatasetID.write
  File "h5py\\_proxy.pyx", line 147, in h5py._proxy.dset_rw
  File "h5py\\_conv.pyx", line 442, in h5py._conv.str2vlen
  File "h5py\\_conv.pyx", line 96, in h5py._conv.generic_converter
  File "h5py\\_conv.pyx", line 247, in h5py._conv.conv_str2vlen
TypeError: Can't implicitly convert non-string objects to strings
```

Is the error handling for this something we want to have on this PR, or should this be handled separately (namely after #350 is merged in)?

bittremieux: I don't think that that will change with the new DepthCharge version, so we can address it here already as well.

Lilferrit:
I added some light error handling such that if the `AnnotatedSpectrumIndex` throws a `TypeError` an additional error message will be thrown pointing the user in the direction of there being an unannotated MGF file in the validation peak path list. I think this provides a nice balance between diagnosing the reason for the error without relying too heavily on the underlying implementation of Depth Charge.

tests/test_integration.py
```diff
@@ -66,15 +89,16 @@ def test_train_and_run(
...
+        "--evaluate",
     ]
 
-    result = run(predict_args)
+    result = run(eval_args)
     assert result.exit_code == 0
     assert output_filename.is_file()
 
     mztab = pyteomics.mztab.MzTab(str(output_filename))
     # Verify that both input peak files are listed in the metadata.
```
bittremieux, line 100: **nitpick:** Comment is no longer correct..

tests/test_integration.py
```diff
@@ -66,15 +89,16 @@ def test_train_and_run(
...
-    for i, filename in enumerate(["small.mgf", "small.mzml"], 1):
+    for i, filename in enumerate(["small.mgf"], 1):
```
bittremieux, line 101: **nitpick:** The for loop could be removed, because only a single file needs to be checked.

CHANGELOG.md
bittremieux, line 1: **todo:** Add the new functionality in the changelog.

bittremieux (changes requested):
A few more final tweaks.

bittremieux (approved):

PR #362, opened 2024-08-02
MzTab Validation

Lilferrit:
Automatically runs mzTab validation via github actions using a compiled jmzTab validator executable.


Lilferrit:
I am not looking for a review yet I just opened this PR to run the test jobs and test the jmzTab validator steps.

tests/mzTab_validation/tiny_model.ckpt
bittremieux, line 1: **thought:** One concern I have with this approach is that it adds a decent amount of bloat to the repository: 37 MB for the model weights and 27 MB for the mzTab validator jar.

The first one might not be necessary. Instead of running Casanovo in an external script, we can just run the integration test, which uses random weights and exports the results to a temporary mzTab file. And then we can call the jar from within the integration test with `subprocess.Popen` or something similar. Bonus benefit: no need to distinguish between different OS platforms (although the implementation was nice).

Adding the jar to the repository probably can't be avoided, unless we host it somewhere else. But this is prone to other issues. To briefly discuss during the meeting.

bittremieux (changes requested):

tests/mzTab_validation/tiny_model.ckpt
Lilferrit: Agreed, I reimplemented the test as an integration test instead of a script. This also had the benefit of eliminating all of the new files except for the jmzTab executable itself.

tests/test_integration.py
```diff
@@ -1,11 +1,14 @@
 import functools
+import subprocess
 from pathlib import Path
 
 import pyteomics.mztab
 from click.testing import CliRunner
 
 from casanovo import casanovo
 
+TEST_DIR = Path(__file__).resolve().parent
```
bittremieux, line 10: **nitpick:** Two blank lines after the imports.

bittremieux (approved):
Nice clean implementation. 👍 One minor style comment.

One more thing that you might investigate a bit more before merging though is whether the validation can still be extracted. I'm thinking something like a separate GitHub action, where it now says "Set up Java" and uses that specific action, that we might create a dedicated action that provides mzTab validation. I haven't done something like that myself yet, so I don't know how it would work or whether it'd be possible.

The immediate benefit is that the jar would not need to be included in the Casanovo repository, but rather in a separate repository that would implement this action. The bigger benefit, however, would be that we'd then be able to easily call this action for any tool that exports to mzTab, whereas now the validation is Casanovo-specific.

bittremieux:
Hmm, question though: the change to export the precursor charge as an integer instead of a float in [9792584](https://github.com/Noble-Lab/casanovo/pull/362/commits/97925848ce2e7f74cb455003cd34f356f576cc4e) doesn't seem to be retained in the final diff. Did that get reverted? And if so, the generated mzTabs should be non-compliant, so the test should fail. Can you double-check that the output is correct and the tests run successfully?

The precursor charge export fix should also be mentioned in the changelog.

Lilferrit:
> Hmm, question though: the change to export the precursor charge as an integer instead of a float in [9792584](https://github.com/Noble-Lab/casanovo/pull/362/commits/97925848ce2e7f74cb455003cd34f356f576cc4e) doesn't seem to be retained in the final diff. Did that get reverted? And if so, the generated mzTabs should be non-compliant, so the test should fail. Can you double-check that the output is correct and the tests run successfully?
> 
> The precursor charge export fix should also be mentioned in the changelog.

I checked `ms_io.py` on the latest comment on this branch and the precursor charge fix is still there. At one point I removed the fix to make sure that the pipeline fails if the output file isn't compliant, but I simply put the fix back when I was done. This probably messed with the git diff. The changelog still has the fix in it as well.

bittremieux:
Yes, you're right. Very weird, the diff doesn't even indicate that `ms_io.py` was changed, nor the changelog.

bittremieux (approved):

PR #365, opened 2024-08-06
Save best model

Lilferrit:
Save model with the lowest validation loss to the checkpoint file "best.ckpt". This is implemented by locking `save_top_k` on the validation loss callback to 1 and eliminating the `save_top_k` option from the Casanovo config.


bittremieux (changes requested):
Looks good. One thing to add is to give a warning to the user that the `save_top_k` setting has been removed. The current [deprecation warning](https://github.com/Noble-Lab/casanovo/blob/main/casanovo/config.py#L18) can be modified to report removal in addition to renaming of settings (e.g. by setting the value to `None` and modifying the warning a bit further down).

Additionally, this should be mentioned in the changelog.

casanovo/config.py
```diff
@@ -96,10 +96,22 @@ def __init__(self, config_file: Optional[str] = None):
                 # Remap deprecated config entries.
                 for old, new in _config_deprecated.items():
                     if old in self._user_config:
-                        self._user_config[new] = self._user_config.pop(old)
+                        warning_msg = (
```
bittremieux, line 99: **nitpick:** I would move this into the `else` to mirror the behavior in the `if`.

casanovo/config.py
```diff
@@ -96,10 +96,22 @@ def __init__(self, config_file: Optional[str] = None):
...
+                            "is no longer in use"
+                        )
+
+                        if new is not None:
+                            self._user_config[new] = self._user_config.pop(old)
+                            warning_msg = (
+                                f"Deprecated config option '{old}' "
+                                f"remapped to '{new}'"
+                            )
+                        else:
```
bittremieux, line 110: **todo:** Can you add a test for this situation [here](https://github.com/Noble-Lab/casanovo/blob/save-best/tests/unit_tests/test_config.py#L42)?

casanovo/config.py
```diff
@@ -96,10 +96,22 @@ def __init__(self, config_file: Optional[str] = None):
...
+                            del self._user_config[old]
+
                         warnings.warn(
-                            f"Deprecated config option '{old}' remapped to "
-                            f"'{new}'",
+                            warning_msg,
```
bittremieux, line 114: **nitpick:** Can be moved onto a single line.

bittremieux (changes requested):

casanovo/config.py
```diff
@@ -96,12 +96,20 @@ def __init__(self, config_file: Optional[str] = None):
...
+                        if new is not None:
+                            self._user_config[new] = self._user_config.pop(old)
+                            warning_msg = (
+                                f"Deprecated config option '{old}' "
+                                f"remapped to '{new}'"
+                            )
+                        else:
+                            del self._user_config[old]
+                            warning_msg = (
+                                f"Depreciated config option '{old}' "
```
bittremieux, line 108: **typo:** Deprecated

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Precursor charges are exported as integers instead of floats in the mzTab output file, in compliance with the mzTab specification.
 
+### Removed
+
+- Removed the `save_top_k` option from the Casanovo config, the model with the lowest validation loss during training will now be saved to a fixed filename `best.ckpt`.
```
bittremieux, line 20: **todo:** Sorry for not catching this earlier, but it should not actually be `best.ckpt`, but rather `<prefix>.best.ckpt` if an output name is specified, as per our discussion [here](https://docs.google.com/document/d/1szkb7l8Obm01UESl8UNj53Dm4W3Y7BG_lO1eHX4H_LA/edit). The final API will still be a bit different, with both a directory and a file root that can be specified, but I think we should already incorporate this to the extent possible in this PR.

tests/unit_tests/test_config.py
```diff
@@ -43,9 +43,18 @@ def test_deprecated(tmp_path, tiny_config):
...
+        # Insert remapped deprecated config option.
         cfg["max_iters"] = 1
         yaml.safe_dump(cfg, f_out)
 
     with pytest.warns(DeprecationWarning):
         Config(filename)
+
+    with open(tiny_config, "r") as f_in, open(filename, "w") as f_out:
+        cfg = yaml.safe_load(f_in)
+        # Insert non-remapped depreciated config option.
```
bittremieux, line 55: **typo:** deprecated

bittremieux (changes requested):

CHANGELOG.md
Lilferrit: Sounds good, I added an option to the `ModelRunner` constructor, and I'll extend the functionality in future PRs.

bittremieux (approved):

PR #372, opened 2024-08-27
File IO command line options revision

Lilferrit:
Removed the `model_save_folder_path` option in favor of the new `--output_root` and `--ouput_dir` command line options as written in the Casanovo File IO specification document, and introduced a new `--overwrite` flag that is required to be set in order to overwrite any output files. The `--validation_peak_path` option is now optional during training. If the `--validation_peak_path` option is not set the `train_peak_path` will be used to validate the model. In this case all spectra files in `train_peak_path` must have peptide annotations.


casanovo/utils.py
```diff
@@ -251,3 +254,34 @@ def log_sequencing_report(
...
+    FileExistsError
+        If matching file name is found in dir
+    """
+    for pattern in file_patterns:
+        comp_pattern = re.compile(pattern)
+        for file in dir.iterdir():
+            if not file.is_file():
+                continue
+
+            if comp_pattern.fullmatch(file.name) is not None:
```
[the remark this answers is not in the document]
bittremieux: Ok, now I see that this is not necessary for the log and mztab check, but it is for the checkpoints check. However, maybe [glob](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) is a better fit.

casanovo/denovo/model_runner.py
```diff
@@ -59,20 +63,35 @@ def __init__(
...
+                    re.escape(output_rootname + ".") + pattern
+                    for pattern in patterns
+                ]
+            utils.check_dir(output_dir, patterns)
 
         # Configure checkpoints.
         self.callbacks = [
             ModelCheckpoint(
-                dirpath=config.model_save_folder_path,
+                dirpath=output_dir,
```
bittremieux, line 85: **todo:** `output_dir` could still be `None` here, in which case the previous file existing checks have not been done. However, this will use the default Lightning directory, and so files can still get overwritten. I think we need to explicitly set `output_dir` earlier in the code.

tests/test_integration.py
```diff
@@ -29,28 +30,40 @@ def test_train_and_run(
...
-    best_model = tmp_path / "best.ckpt"
+    model_file = tmp_path / "train.epoch=19-step=20.ckpt"
+    best_model = tmp_path / "train.best.ckpt"
     assert result.exit_code == 0
     assert model_file.exists()
     assert best_model.exists()
 
-    # Finally try predicting:
-    output_filename = tmp_path / "test.mztab"
+    # Check that re-running train fails due to no overwrite
```
bittremieux, line 47: **suggestion:** I don't think we need to cover this in the integration test, it's already verified in `test_runner` where the relevant error is actually thrown.

casanovo/utils.py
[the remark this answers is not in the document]
Lilferrit: I took a look at glob and it looks like the regex that glob uses is limited compared to the Python re module. For example, I don't think it would be possible to create regex that matches the checkpoint file names exactly with glob. It would be possible to create a pattern (`<output_root>.epoch=*step=*.ckpt` or similar) that matches a superset of checkpoint names, for example it would match `<output_root>.epoch=foo-step=bar.ckpt`. Maybe the decrease in code complexity would be worth the tradeoff in this case though.

casanovo/casanovo.py
```diff
@@ -90,6 +95,15 @@ def __init__(self, *args, **kwargs) -> None:
                 ),
                 default="info",
             ),
+            click.Option(
+                ("-d", "--overwrite"),
+                help="""
```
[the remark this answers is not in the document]
Lilferrit: Maybe -f standing for something like "force" would be more appropriate? The -o shorthand is already taken by `--output_root`.

casanovo/casanovo.py
```diff
@@ -90,6 +95,15 @@ def __init__(self, *args, **kwargs) -> None:
                 ),
                 default="info",
             ),
+            click.Option(
+                ("-d", "--overwrite"),
```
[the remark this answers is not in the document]
Lilferrit: Maybe -f standing for something like "force" would be more appropriate? The -o shorthand is already taken by --output_root.

casanovo/utils.py
```diff
@@ -251,3 +254,34 @@ def log_sequencing_report(
...
+    file_patterns : Iterable[re.Pattern[str]]
+        File name re patterns to test file names against
+
+    Raises
+    ------
+    FileExistsError
+        If matching file name is found in dir
+    """
+    for pattern in file_patterns:
+        comp_pattern = re.compile(pattern)
```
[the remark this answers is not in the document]
Lilferrit: My understanding is that the regex would be tested against each of the files in the target directory, I may be misunderstanding something though.

casanovo/denovo/model_runner.py
```diff
@@ -59,20 +63,35 @@ def __init__(
         self.loaders = None
         self.writer = None
 
-        best_filename = "best"
+        filenames = ("{epoch}-{step}", "best")
```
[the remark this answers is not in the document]
Lilferrit: Good call, thanks!

casanovo/utils.py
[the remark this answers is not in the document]
bittremieux: Yes, it's indeed a bit more limited functionality, but sufficient for what we want to do imo.

casanovo/casanovo.py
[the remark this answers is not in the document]
bittremieux: Yes, "force" makes more sense. The long version needs to be made consistent with that as well then.

casanovo/utils.py
```diff
@@ -251,3 +254,28 @@ def log_sequencing_report(
...
+    Raises
+    ------
+    FileExistsError
+        If matching file name is found in dir
+    """
+    for pattern in file_patterns:
+        matches = list(dir.glob(pattern))
+        if len(matches) > 0:
+            raise FileExistsError(
+                f"File {matches[0].name} already exists in {dir} "
```
bittremieux, line 279: **thought:** This might be a slightly misleading error message. The user could remove the file that's mentioned, but there could be multiple files matching the pattern, which are not explicitly mentioned. Rather than reporting all files, which might result in an unwieldy error message, maybe just report the pattern instead?

```
if next(dir.glob(pattern), None) is None:
    raise FileExistsError(...)
```

This also avoids having to consume the whole generator (which admittedly doesn't really matter that much, even if there are dozens to hundreds of checkpoints matching).

bittremieux (changes requested):

casanovo/casanovo.py
```diff
@@ -489,6 +522,44 @@ def _get_model_weights(cache_dir: Path) -> str:
...
+    verbosity : str
+        The verbosity level for logging.
+
+    Returns:
+    --------
+    Tuple[Path, str]
+        The output directory and the base name for log and results files (if
+        applicable).
+    """
+    output_dir = Path(output_dir) if output_dir is not None else Path.cwd()
```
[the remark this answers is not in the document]
Lilferrit: IMO for a non-existing path it would be best to throw an exception and let the user create the path if that's where they actually wanted to write the output files to, or at the very least pop a warning if a new path is created.

casanovo/casanovo.py
```diff
@@ -489,6 +522,44 @@ def _get_model_weights(cache_dir: Path) -> str:
...
+    output_base_name = (
+        None if output_root is None else (output_dir / output_root)
+    )
+    utils.check_dir(output_dir, file_patterns)
+    output = setup_logging(output_base_name, verbosity)
+    return output, output_dir
+
```
Lilferrit, line 561: It might be better to pop a warning here instead, not sure.

casanovo/utils.py
Lilferrit: Good call!

casanovo/casanovo.py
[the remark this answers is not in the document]
bittremieux: I think an exception is maybe a bit excessive fora non-existing path. You could envision running Casanovo in a script with slightly different input every time and writing to different subfolders. Instead, creating the path and issuing a warning seems best.

casanovo/casanovo.py
bittremieux: I'm not sure what this comment refers to. What do we need to warn about?

casanovo/casanovo.py
```diff
@@ -259,21 +289,11 @@ def setup_logging(
...
-
-    Return
-    ------
-    output : Path
-        The output file path.
     """
-    if output is None:
-        output = f"casanovo_{datetime.datetime.now().strftime('%Y%m%d%H%M%S')}"
-
-    output = Path(output).expanduser().resolve()
```
bittremieux, line 275: **question:** The `expanduser` and `resolve` are not reflected in the new functionality at the bottom of this file. Is this still being done? Explicitly expanding the user was necessary to ensure cross-platform functionality (e.g. paths with `~` can be resolved natively in Unix but not in Windows, which is why we explicitly did it ourselves). Please verify that this functionality is retained.

casanovo/denovo/model_runner.py
```diff
@@ -59,20 +70,39 @@ def __init__(
         self.loaders = None
         self.writer = None
 
-        best_filename = "best"
-        if output_rootname is not None:
-            best_filename = f"{output_rootname}.{best_filename}"
+        if output_dir is None:
+            self.callbacks = []
+            warnings.warn(
```
bittremieux, line 75: **question:** Any benefit over emitting a warning vs logging with the warning level. I think the latter seems more relevant here.

casanovo/casanovo.py
Lilferrit: This comment was supposed to be on the line `raise FileNotFoundError(`, my bad.

casanovo/casanovo.py
```diff
@@ -489,6 +507,58 @@ def _get_model_weights(cache_dir: Path) -> str:
...
+            f"casanovo_{datetime.datetime.now().strftime('%Y%m%d%H%M%S')}"
+        )
+
+    if output_dir is None:
+        output_path = Path.cwd()
+    else:
+        output_path = Path(output_dir).expanduser().resolve()
+        if not output_path.is_dir():
+            output_path.mkdir(parents=True)
+            warnings.warn(
```
bittremieux, line 550: **todo:** This should be `logging.warn(...)` instead.

Conceptually you can consider `warnings.warn` messages for _developers_ and `logging.xxx` messages for users.

casanovo/casanovo.py
```diff
@@ -154,18 +168,25 @@ def sequence(
     to sequence peptides. If evaluate is set to True PEAK_PATH must be
     one or more annotated MGF file.
     """
-    output = setup_logging(output, verbosity)
-    config, model = setup_model(model, config, output, False)
+    output_path, output_root = _setup_output(
+        output_dir, output_root, force_overwrite, verbosity
+    )
+    utils.check_dir_file_exists(output_path, f"{output_root}.mztab")
+    config, model = setup_model(model, config, output_dir, output_root, False)
```
bittremieux, line 175: **todo:** Shouldn't this use the resolved `output_path` instead of `output_dir`, which can be None?

casanovo/casanovo.py
```diff
@@ -186,31 +207,40 @@ def sequence(
...
 
     TRAIN_PEAK_PATH must be one or more annoated MGF files, such as those
     provided by MassIVE-KB, from which to train a new Casnovo model.
     """
-    output = setup_logging(output, verbosity)
-    config, model = setup_model(model, config, output, True)
+    output_path, output_root = _setup_output(
+        output_dir, output_root, force_overwrite, verbosity
+    )
+    config, model = setup_model(model, config, output_path, output_root, True)
```
bittremieux, line 232: **note:** Whereas here `output_path` is correctly used.

casanovo/casanovo.py
bittremieux: **todo:** I also that the docstring for `setup_model` is outdated (not shown here in the diff). Please update that as well.

bittremieux (changes requested):

casanovo/casanovo.py
```diff
@@ -489,6 +527,58 @@ def _get_model_weights(cache_dir: Path) -> str:
...
+        )
+
+    if output_dir is None:
+        output_path = Path.cwd()
+    else:
+        output_path = Path(output_dir).expanduser().resolve()
+        if not output_path.is_dir():
+            output_path.mkdir(parents=True)
+            logger.warning(
+                f"Target output directory {output_dir} does not exists, "
```
bittremieux, line 571: **nitpick:** Use `%s` instead of f-string for logging.

bittremieux (approved):
Looks good, just one trivial suggestions before merging.

PR #374, opened 2024-09-06
Tensorboard io

Lilferrit:
TensorBoard summary writer now writes to a subdirectory of `--output_dir`, `<output_dir>/tensorboard`. The config option `tb_summarywriter` has been redefined to be a boolean, and if set to true the TensorBoard summary will be written to the output directory.

I am marking this PR as a draft until `console-file-io` get's merged into dev, as this branch is based on `console-file-io` and is dependent on changes introduced in that branch.


Lilferrit:
Alternatively we could introduce a `--tensorboard` or similar command line option, and eliminate the `tb_summarywriter` config option entirely.

bittremieux (approved):
Can be merged after #372 has been completed.

bittremieux:
> Alternatively we could introduce a `--tensorboard` or similar command line option, and eliminate the `tb_summarywriter` config option entirely.

No, the current solution is better. This is not a very important setting that we want to expose on the command line.

PR #376, opened 2024-09-16
Log optimizer and training metrics to CSV file

Lilferrit:
Implemented logging of optimizer (e.g. learning rate and weight decay) and training (e.g. validation and train loss) to a CSV file using `lightning.pytorch.loggers.CSVLogger`. Also added an additional callback to log optimizer metrics.


casanovo/denovo/model_runner.py
```diff
@@ -105,6 +108,10 @@ def __init__(
                 filename=best_filename,
                 enable_version_counter=False,
             ),
+            LearningRateMonitor(
+                log_momentum=True,
```
bittremieux, line 112: **nitpick:** I think these two can be combined in a single line without it being too wide.

casanovo/denovo/model_runner.py
```diff
@@ -255,7 +262,36 @@ def initialize_trainer(self, train: bool) -> None:
                 strategy=self._get_strategy(),
                 val_check_interval=self.config.val_check_interval,
                 check_val_every_n_epoch=None,
+                log_every_n_steps=1,
```
bittremieux, line 265: **question:** Shouldn't this use the logging interval from the config?

casanovo/denovo/model_runner.py
```diff
@@ -255,7 +262,36 @@ def initialize_trainer(self, train: bool) -> None:
...
             )
+
+            if self.config._user_config.get("log_metrics"):
```
bittremieux, line 268: **question:** Why are you using `_user_config`? As indicated by the underscore, this is supposed to be private. Instead, it should just be `self.config["log_metrics"]`.

casanovo/denovo/model_runner.py
```diff
@@ -255,7 +262,36 @@ def initialize_trainer(self, train: bool) -> None:
...
+                if not self.output_dir:
+                    logger.warning(
+                        "Output directory not set in model runner. "
+                        "No loss file will be created."
+                    )
+                else:
+                    CSV_LOG_SUB_DIR = "csv_logs"
```
bittremieux, line 275: **question:** Not sure why this is all caps.

casanovo/denovo/model_runner.py
```diff
@@ -255,7 +262,36 @@ def initialize_trainer(self, train: bool) -> None:
...
+                        )
+
+                    additional_cfg.update(
+                        {
+                            "logger": lightning.pytorch.loggers.CSVLogger(
+                                self.output_dir,
+                                version=CSV_LOG_SUB_DIR,
+                                name=None,
+                            ),
+                            "log_every_n_steps": self.config._user_config.get(
```
bittremieux, line 289: **todo:** Here also improper way to retrieve a config value.

casanovo/denovo/model_runner.py
```diff
@@ -255,7 +262,36 @@ def initialize_trainer(self, train: bool) -> None:
...
+                        utils.check_dir_file_exists(
+                            self.output_dir,
+                            CSV_LOG_SUB_DIR,
+                        )
+
+                    additional_cfg.update(
+                        {
+                            "logger": lightning.pytorch.loggers.CSVLogger(
+                                self.output_dir,
+                                version=CSV_LOG_SUB_DIR,
```
bittremieux, line 286: **question:** Shouldn't `name` be set instead of `version`?

bittremieux (changes requested):

casanovo/denovo/model_runner.py
Lilferrit: When first implemented this I used `self.config["log_metrics"]`, but I ran into an issue where the default value was being retrieved instead of the actual value in the config file. I'm not running into this issue now though so I'll go ahead and use the config objects item retrieval instead.

casanovo/denovo/model_runner.py
Lilferrit: It's a constant string I use more than once, but since it's not defined at the module level I'll go ahead and rename it.

casanovo/denovo/model_runner.py
Lilferrit: If you don't define a version PyLightning will automatically create version sub directories under `save_dir`. I tried setting `version` to an empty string and it still created the version sub directories (if I had to guess they have something like `if not version:` somewhere in their code). Since, in their own words, logs will be saved to `os.path.join(save_dir, name, version)` (https://lightning.ai/docs/pytorch/stable/extensions/generated/lightning.pytorch.loggers.CSVLogger.html), this is a work-around I came up with to save the logs directly to `<output_dir>/csv_logs`.

casanovo/denovo/model_runner.py
Lilferrit: Yep, updated.

casanovo/casanovo.py
```diff
@@ -192,7 +192,7 @@ def sequence(
 
         runner.predict(
             peak_path,
-            str((output_path / output_root).with_suffix(".mztab")),
+            str((output_path / output_root_name).with_suffix(".mztab")),
```
Lilferrit, line 195: Added this as an additional bug fix.

casanovo/denovo/model_runner.py
```diff
@@ -255,7 +259,36 @@ def initialize_trainer(self, train: bool) -> None:
...
+                if not self.output_dir:
+                    logger.warning(
+                        "Output directory not set in model runner. "
+                        "No loss file will be created."
+                    )
+                else:
+                    csv_log_dir = "csv_logs"
+                    if self.overwrite_ckpt_check:
+                        utils.check_dir_file_exists(
+                            self.output_dir,
```
bittremieux, line 275: **nitpick:** This probably also doesn't need to be split over two lines.

bittremieux (approved):

bittremieux:
Looks good. Please update the CHANGELOG before merging.

Issue #379, opened 2024-09-17
Intermittent circular import error


Lilferrit:
The current `dev` branch has an intermittent circular import bug. A simple way to reproduce this bug is to run the command `pytest tests\unit_tests\test_runner.py` (this error doesn't occur in my local environment if all unit tests are ran at once), which then gives the output:

```
ImportError while importing test module 'C:\Users\gavin\OneDrive\Documents\casanovo\tests\unit_tests\test_runner.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
D:\anaconda3\envs\casanovo_env\lib\importlib\__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests\unit_tests\test_runner.py:8: in <module>
    from casanovo.config import Config
casanovo\config.py:11: in <module>
    from . import utils
casanovo\utils.py:18: in <module>
    from .data.ms_io import PepSpecMatch
casanovo\data\ms_io.py:15: in <module>
    from ..config import Config
E   ImportError: cannot import name 'Config' from partially initialized module 'casanovo.config' (most likely due to a circular import) (C:\Users\gavin\OneDrive\Documents\casanovo\casanovo\config.py)
```

This issue is resolved if the `PepSpecMatch` data class is moved to it's own file.

PR #380, opened 2024-09-17
Eval metrics and circular import bug fix.

### LINKED ISSUE #378: `log_metrics` skipped spectra incorrect behavior
The current implementationof `ModelRunner.log_metrics` on the `dev` branch does not handle skipped spectra properly, leading to the peptide precision score being less than it's actual value. A simple test case to reproduce this incorrect behavior is:

```python
def test_eval_bug(monkeypatch, tiny_config):
    TEST_EPSILON = 10 ** -5

    def get_mock_index(psm_list):
        mock_test_index = unittest.mock.MagicMock()
        mock_test_index.__enter__.return_value = mock_test_index
        mock_test_index.__exit__.return_value = False
        mock_test_index.n_spectra = len(psm_list)
        mock_test_index.get_spectrum_id = lambda idx: psm_list[idx].spectrum_id

        mock_spectra = [
            (None, None, None, None, curr_psm.sequence)
            for curr_psm in psm_list
        ]
        mock_test_index.__getitem__.side_effect = lambda idx: mock_spectra[idx]
        return mock_test_index

    def get_mock_psm(sequence, spectrum_id):
        return PepSpecMatch(
            sequence=sequence,
            spectrum_id=spectrum_id,
            peptide_score=None,
            charge=None,
            exp_mz=None,
            aa_scores=None,
            calc_mz=None,
        )

    with monkeypatch.context() as ctx:
        mock_logger = unittest.mock.MagicMock()
        ctx.setattr("casanovo.denovo.model_runner.logger", mock_logger)

        with ModelRunner(Config(tiny_config)) as runner:
            runner.writer = unittest.mock.MagicMock()

            # Test skipped spectra
            act_psms = [
                get_mock_psm("PEP", ("foo", "index=1")),
                get_mock_psm("PET", ("foo", "index=2")),
                get_mock_psm("PEI", ("foo", "index=3")),
                get_mock_psm("PEG", ("foo", "index=4")),
                get_mock_psm("PEA", ("foo", "index=5")),
            ]

            infer_psms = [
                get_mock_psm("PET", ("foo", "index=2")),
                get_mock_psm("PEI", ("foo", "index=3")),
                get_mock_psm("PEG", ("foo", "index=4")),
                get_mock_psm("PEA", ("foo", "index=5")),
            ]

            runner.writer.psms = infer_psms
            mock_index = get_mock_index(act_psms)
            runner.log_metrics(mock_index)
            pep_precision = mock_logger.info.call_args_list[-2][0][1]
            aa_precision = mock_logger.info.call_args_list[-1][0][1]
            
            print(f"AA precision: {aa_precision}")
            print(f"Peptide precision: {pep_precision}")

            assert abs(pep_precision - 100 * (4 / 5)) < TEST_EPSILON
            assert abs(aa_precision - 100 * (12 / 13)) < TEST_EPSILON
``` 

The output of the test case (which fails) is:

```
AA precision: 66.6666666111111
Peptide precision: 0.0
```

In this case, the values reported for peptide precision and amino acid precision are both incorrect. The likely cause of this bug in the current implementation of `log_metrics` is here:

```python
model_output = [psm.sequence for psm in self.writer.psms]
spectrum_annotations = [
    test_index[i][4] for i in range(test_index.n_spectra)
]
```

After the first instance of a skipped spectra, the lists `model_output` and `spectrum_annotations` are no longer "aligned" by their spectrum ids.

Lilferrit:
Implemented bug fixes to resolve #378 and #379. Also implemented a unit test for `ModelRunner.log_metrics` to test for future incorrect behavior.


casanovo/denovo/model_runner.py
```diff
@@ -163,18 +163,35 @@ def log_metrics(self, test_index: AnnotatedSpectrumIndex) -> None:
...
+            for i in range(t_ind.n_spectra):
+                if curr_psm is None:
+                    break
+
+                spectrum_annotations.append(t_ind[i][4])
+                if curr_psm.spectrum_id == t_ind.get_spectrum_id(i):
+                    model_output.append(curr_psm.sequence)
+                    curr_psm = next(psms, None)
+                else:
+                    model_output.append("")
```
Lilferrit, line 181: **question**: Is this the correct behavior for handling skipped spectra? I'm not sure if skipped spectra should be ignored entirely or yield a peptide precision of 0.

wsnoble: I think they should be treated as incorrect predictions with scores of -1.

bittremieux: I agree. `None` might be a better value instead of an empty string.

bittremieux (changes requested):

tests/unit_tests/test_runner.py
```diff
@@ -1,12 +1,14 @@
...
 
+import unittest.mock
 from pathlib import Path
 
 import pytest
 import torch
 
 from casanovo.config import Config
 from casanovo.denovo.model_runner import ModelRunner
+from casanovo.data.ms_io import PepSpecMatch
```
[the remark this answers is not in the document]
Lilferrit: Good catch thanks! It technically works because `ms_io` imports PepSpecMatch so I didn't catch this during testing.

casanovo/denovo/model_runner.py
Lilferrit: Using `None` causes `aa_match_batch` to throw an exception. Should `aa_match_batch` be updated to handle `None` or should we just go with the empty string?

bittremieux: I'd suggest to update `aa_match_batch` to handle `None`, so that we have a clear semantic indication of the spectrum being skipped.

casanovo/denovo/model_runner.py
```diff
@@ -163,18 +163,28 @@ def log_metrics(self, test_index: AnnotatedSpectrumIndex) -> None:
...
+        with test_index as t_ind:
+            for true_idx in range(t_ind.n_spectra):
+                seq_true.append(t_ind[true_idx][4])
+                if pred_idx < len(self.writer.psms) and self.writer.psms[
+                    pred_idx
+                ].spectrum_id == t_ind.get_spectrum_id(true_idx):
+                    seq_pred.append(self.writer.psms[pred_idx].sequence)
+                    pred_idx += 1
+                else:
+                    seq_pred.append("")
```
bittremieux, line 179: **todo:** Use `None` for skipped spectra.

casanovo/denovo/model_runner.py
```diff
@@ -163,18 +163,28 @@ def log_metrics(self, test_index: AnnotatedSpectrumIndex) -> None:
...
-            test_index[i][4] for i in range(test_index.n_spectra)
-        ]
+        seq_pred = []
+        seq_true = []
+        pred_idx = 0
+
+        with test_index as t_ind:
+            for true_idx in range(t_ind.n_spectra):
+                seq_true.append(t_ind[true_idx][4])
+                if pred_idx < len(self.writer.psms) and self.writer.psms[
```
bittremieux, line 173: **todo:** I think this might not fully work if there are gaps of multiple skipped spectra in a row.

tests/unit_tests/test_runner.py
```diff
@@ -282,3 +284,151 @@ def test_evaluate(
...
+            runner.writer.psms = infer_psms
+            mock_index = get_mock_index(act_psms)
+            runner.log_metrics(mock_index)
+
+            pep_precision = mock_logger.info.call_args_list[-2][0][1]
+            aa_precision = mock_logger.info.call_args_list[-1][0][1]
+            assert pep_precision == pytest.approx(100 * (1 / 2))
+            assert aa_precision == pytest.approx(100 * (5 / 6))
+
+            # Test skipped spectra
```
bittremieux, line 362: **todo:** Add a test case where there are multiple skipped spectra in a row.

bittremieux (changes requested):
Looks good. One more thing to test/verify: multiple skipped spectra in a row. And then I think switching to `None` for skipped spectra instead of an empty string, and updating `aa_match_batch` to handle this, improves clarity.

casanovo/denovo/model_runner.py
Lilferrit: I added a test case with a gap of multiple skipped spectra in the middle of the spectra annotations and one where the gap ends at the end of the spectra annotations, and everything seems to be working as expected.

Lilferrit:
The updated `aa_match_batch` now handles amino acid precision for skipped spectra differently. For example, say for the ground truth peptides PEP and PET the model only (correctly) infers PEP and skips PET, the amino acid precision will now be 1 or 100%. Before, the first amino acid in any skipped spectrum would be treated as an incorrect inference, while the rest wouldn't be counted, so the peptide precision for the last case would be .75, and I'm not sure if that was correct.

casanovo/denovo/evaluate.py
```diff
@@ -225,8 +225,14 @@ def aa_match_batch(
         # Split peptides into individual AAs if necessary.
         if isinstance(peptide1, str):
             peptide1 = re.split(r"(?<=.)(?=[A-Z])", peptide1)
+        elif peptide1 is None:
```
bittremieux, line 228: **suggestion:** I'd move this check first, and then have the string check be part of the `elif`.

bittremieux (commented):
Looks good. Let's finish the discussion on Slack how to account for skipped spectra when calculating amino acid precision before merging.

I realized a situation where the evaluation might fail though: If we have multiple predictions per spectrum (i.e. `top_match` in the config > 1). I think that this might have failed in the previous implementation as well though. And it's not super obvious how to handle this situation (or if we should—but maybe at least a check/warning).

Lilferrit:
> I realized a situation where the evaluation might fail though: If we have multiple predictions per spectrum (i.e. top_match in the config > 1). I think that this might have failed in the previous implementation as well though. And it's not super obvious how to handle this situation (or if we should—but maybe at least a check/warning).

To me the most intuitive way to handle this is to only evaluate the highest confidence PSM for each spectrum. If I'm understanding everything correctly the current implementation just matches whatever PSM happens to be first in `writer.psms` to the ground truth spectra, which might be ok (?), but as you said I'm also not sure if it's worth implementing special handling for the case where `top_match` > 1. For now I'll just add a warning that pops if `top_match` is not one.

wsnoble:
> To me the most intuitive way to handle this is to only evaluate the highest confidence PSM for each spectrum. 

I agree with this.

tests/unit_tests/test_runner.py
```diff
@@ -282,3 +285,223 @@ def test_evaluate(
...
+        ) as runner:
+            runner.train([mgf_small], [mgf_small])
+
+    assert not curr_model_path.is_file()
+    assert not best_model_path.is_file()
+    assert not tb_path.is_dir()
+    assert csv_path.is_dir()
+
+
+def test_log_metrics(monkeypatch, tiny_config):
```
Lilferrit, line 328: The git history here got messed up during a rebase, the only new addition here is now each log_metrics test case also tests aa_recall in addition to aa_precison.

casanovo/denovo/model_runner.py
```diff
@@ -255,7 +277,34 @@ def initialize_trainer(self, train: bool) -> None:
...
+                        )
+
+                    additional_cfg.update(
+                        {
+                            "logger": lightning.pytorch.loggers.CSVLogger(
+                                self.output_dir,
+                                version=csv_log_dir,
+                                name=None,
+                            ),
+                            "log_every_n_steps": self.config.log_every_n_steps,
```
Lilferrit, line 304: Had to fix a bug here that for some reason did not appear on the dev branch.

casanovo/config.yaml
```diff
@@ -42,6 +42,10 @@ random_seed: 454
 n_log: 1
 # Whether to create tensorboard directory
 tb_summarywriter: false
+# Whether to create csv_logs directory
+log_metrics: false
+# How often to log optimizer parameters in steps
+log_every_n_steps: 50
```
Lilferrit, line 48: History also got messed up here during a rebase

casanovo/config.py
```diff
@@ -65,6 +65,8 @@ class Config:
         residues=dict,
         n_log=int,
         tb_summarywriter=bool,
+        log_metrics=bool,
+        log_every_n_steps=int,
```
Lilferrit, line 69: Same thing, rebase messed up the history

casanovo/casanovo.py
```diff
@@ -192,7 +192,7 @@ def sequence(
 
         runner.predict(
             peak_path,
-            str((output_path / output_root).with_suffix(".mztab")),
+            str((output_path / output_root_name).with_suffix(".mztab")),
```
Lilferrit, line 195: rebase also messed up the history here

tests/unit_tests/test_runner.py
```diff
@@ -282,3 +285,223 @@ def test_evaluate(
             )
 
     result_file.unlink()
+
+
+def test_metrics_logging(tmp_path, mgf_small, tiny_config):
```
Lilferrit, line 290: A rebase messed up the git history here, this test is already on `dev`.

tests/conftest.py
```diff
@@ -224,7 +224,9 @@ def tiny_config(tmp_path):
         "devices": None,
         "random_seed": 454,
         "n_log": 1,
-        "tb_summarywriter": None,
+        "tb_summarywriter": False,
+        "log_metrics": False,
+        "log_every_n_steps": 50,
```
Lilferrit, line 229: rebase history error, these changes are already on `dev`.

CHANGELOG.md
```diff
@@ -10,12 +10,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 - During training, model checkpoints will be saved at the end of each training epoch in addition to the checkpoints saved at the end of every validation run.
 - Besides as a local file, model weights can be specified from a URL. Upon initial download, the weights file is cached for future re-use.
+- Training and optimizer metrics can now be logged to a CSV file by setting the `log_metrics` config file option to true - the CSV file will be written to under a sub-directory of the output directory named `csv_logs`.
 
 ### Changed
 
 - Removed the `evaluate` sub-command, and all model evaluation functionality has been moved to the `sequence` command using the new `--evaluate` flag.
 - The `--output` option has been split into two options, `--output_dir` and `--output_root`.
 - The `--validation_peak_path` is now optional when training; if `--validation_peak_path` is not set then the `train_peak_path` will also be used for validation.
+- The `tb_summarywriter` config option is now a boolean config option, and if set to true the TensorBoard summary will be written to a sub-directory of the output directory named `tensorboard`.
```
Lilferrit, line 20: Same thing, rebase messed up the history on this file.

casanovo/denovo/evaluate.py
```diff
@@ -161,7 +161,12 @@ def aa_match(
     pep_match : bool
         Boolean flag to indicate whether the two peptide sequences fully match.
     """
-    if mode == "best":
+    if peptide1 is None and peptide2 is None:
+        return np.array([], dtype=bool), False
```
bittremieux, line 165: **todo:** Can we add a quick test for this? Mainly to boost test coverage.

casanovo/denovo/evaluate.py
```diff
@@ -161,7 +161,12 @@ def aa_match(
...
+    elif (peptide1 is None) != (peptide2 is None):
```
bittremieux, line 166: **todo:** Slightly weird use of `!=`. Why not just use `or` instead, that's more legible. Then also the parentheses are not needed.

casanovo/denovo/evaluate.py
```diff
@@ -161,7 +161,12 @@ def aa_match(
...
+        peptide = peptide1 if peptide2 is None else peptide2
+        return np.array([False] * len(peptide)), False
```
bittremieux, line 168: **todo:** This can be slightly more efficient.

Now: You create a list of length 1, then create a new list of `len(pep)` elements, and then convert the list to an array.
Instead: `np.zeros(len(peptide), dtype=bool)` would directly create an array of the right size.

casanovo/denovo/evaluate.py
bittremieux, line 165: **nitpick:** Maybe `np.empty(0, dtype=bool)`  is slightly better.

casanovo/denovo/evaluate.py
Lilferrit: I added this case before I added the `peptide1 is None and peptide2 is None`, so I was basically using the != as an XOR operator. I agree though that the or is definitely a lot more readable now that the first case is  `peptide1 is None and peptide2 is None`.

bittremieux (approved):

PR #391, opened 2024-10-10
Casanovo Nextflow Workflow Documentation

Lilferrit:
Added a Read the Docs page for the Casanovo Nextflow workflow.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -17,6 +17,7 @@ maxdepth: 1
 Getting Started <getting_started.md>
 File Formats <file_formats.md>
 Command Line Interface <cli.rst>
+Casanovo Nextflow Workflow <nextflow.md>
 FAQs <faq.md>
 Contributing <CONTRIBUTING.md>
 Code of Conduct <CODE_OF_CONDUCT.md>
--- a/docs/index.md
+++ b/docs/index.md
@@ -17,7 +17,7 @@ maxdepth: 1
 Getting Started <getting_started.md>
 File Formats <file_formats.md>
 Command Line Interface <cli.rst>
-Casanovo Nextflow Workflow <nextflow.md>
+Nextflow Workflow <nextflow.md>
 FAQs <faq.md>
 Contributing <CONTRIBUTING.md>
 Code of Conduct <CODE_OF_CONDUCT.md>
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -17,6 +17,7 @@ maxdepth: 1
 Getting Started <getting_started.md>
 File Formats <file_formats.md>
 Command Line Interface <cli.rst>
+Casanovo Nextflow Workflow <nextflow.md>
```
bittremieux, line 20: **thought:** Not sure whether the "Casanovo" prefix is needed, because it could be used for almost all pages then.

docs/nextflow.md
bittremieux, line 1: **todo:** I think we need a bit more information here. Why is this useful, what does it do (msConvert + Casanovo + Limelight), which systems (local and remote). We don't need to replicate the external documentation, but should at least provide some more background info.

Lilferrit: Sounds good, I added some more detail to the Nextflow docs page. Let me know if anything needs to be added/changed.

docs/nextflow.md
```diff
@@ -0,0 +1,6 @@
+# Casanovo Nextflow Workflow
+
+To simplify the process of setting up and running Casanovo, a dedicated [Nextflow](https://www.nextflow.io/) workflow is available.
+In addition to automatic the installation of Casanovo and its dependencies, the Casanovo Nextflow workflow provides an automated mass spectrometry data pipeline that converts input data files to a Casanovo compatible format using [msconvert](https://bioinformaticshome.com/tools/proteomics/descriptions/msconvert.html#gsc.tab=0), infers peptide sequences using Casanovo, and (optionally) uploads the results to [Limelight](https://limelight-ms.org/) - a platform for sharing and visualizing proteomics results.
```
bittremieux, line 4: **typo:** "automatic the installation" -> "automate" I think? Although it's twice "automate" in close succession, so small rephrasing recommended. The first one "simplify"?

bittremieux, line 4: **todo:** Official link for msConvert is: https://proteowizard.sourceforge.io/tools/msconvert.html

bittremieux (approved):
Looks good, just two minor changes.

docs/nextflow.md
bittremieux: This still needs to be fixed I think.

PR #392, opened 2024-10-14
Switched to Geometric Mean for Peptide Level Scores

Lilferrit:
Switched the peptide level score calculation from the arithmetic mean to the geometric mean of amino acid scores.

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -1076,7 +1076,7 @@ def _aa_pep_score(
     peptide_score : float
         The peptide score.
     """
-    peptide_score = np.mean(aa_scores)
+    peptide_score = np.exp(np.mean(np.log(aa_scores)))
     aa_scores = (aa_scores + peptide_score) / 2
     if not fits_precursor_mz:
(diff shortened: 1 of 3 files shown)


casanovo/denovo/model.py
```diff
@@ -1076,7 +1076,7 @@ def _aa_pep_score(
     peptide_score : float
         The peptide score.
     """
-    peptide_score = np.mean(aa_scores)
+    peptide_score = np.exp(np.mean(np.log(aa_scores)))
```
Lilferrit, line 1079: I'm not sure if it's strictly necessary to do this calculation in log space for the sequence lengths we'll be dealing with, but I implemented it this way just to reduce any possible rounding errors.

bittremieux (approved):
Looks good.

Two minor suggestions:
- Maybe add another test that doesn't include a `0` amino acid score, because of the resulting `-inf` log and zero geometric mean. This is an edge case anyway, because real amino acid scores will likely never be truly zero.
- Update the CHANGELOG.

PR #399, opened 2024-11-14
Experimental vs Calculated MZ Bug Fix

Lilferrit:
As pointed out by Isha, the `calc_mz` and `precursor_mz` values are currently erroneously swapped when passing PSMs to the output writer.


bittremieux:
Already fixed in #325.

PR #406, opened 2024-12-02
Fix Log Entry to Config File Bug

Lilferrit:
Fixed a bug that would write the log entries to the config file instead of the log file when running the `configure` command.


casanovo/casanovo.py
```diff
@@ -337,21 +336,46 @@ def version() -> None:
 
 
 @main.command()
+@click.option(
```
bittremieux, line 339: **question:** Could we maybe make two sets of shared commands? One that has these three shared options related to output files, and the other with the remaining options that are in the current `_SharedParams`. Then we don't need to duplicate the first ones and combine those two sets of shared commands where necessary. That would avoid having to redefine those output options here.

casanovo/casanovo.py
```diff
@@ -337,21 +336,46 @@ def version() -> None:
...
+        output_dir, output_root, force_overwrite, "info"
+    )
+    config_fname = output_root if output_root is not None else "casanovo"
+    config_fname = Path(config_fname).with_suffix(".yaml")
+    if not force_overwrite:
+        utils.check_dir_file_exists(output_path, str(config_fname))
+
+    config_path = str(output_path / config_fname)
+    Config.copy_default(config_path)
+    logger.info(f"Wrote {config_path}\n")
```
bittremieux, line 378: **nitpick:** Is the newline at the end necessary?

bittremieux (commented):
Looks good, just some considerations.

As you mentioned on Slack, I think it would also be good to add an integration test that checks whether the produced config file is actually compatible (i.e. you can run Casanovo with it).

casanovo/casanovo.py
Lilferrit: Sounds good, I split the file IO commands into a different class to avoid redefining them.

Lilferrit:
Sounds good, I added additional test cases for the `configure` sub-command as well splitting the file IO options into a different class.

bittremieux (approved):

PR #421, opened 2024-12-17
Upgrade dependencies with security vulnerabilities

bittremieux:


wsnoble (approved):
I am not sure what kind of testing we need to do here.  The changes are super small (just updating two version numbers).

PR #431, opened 2025-02-10
Log Average Throughput and Inference Run Time

Lilferrit:
Introduced end of run logging functionality for throughput and total inference run time. Similar functionality could be implemented using `lightning.pytorch.callbacks.ThroughputMonitor` for less code complexity, however this would log the throughput to the PyLightning logs (e.g. the tensorboard or CSV logs) and I'm not sure if this would be desirable behavior.

### DIFF
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -181,7 +181,8 @@ def sequence(
     config, model = setup_model(
         model, config, output_path, output_root_name, False
     )
-    start_time = time.time()
+    start_time = time.perf_counter()
+    utils.log_intro_report(start_time=start_time)
     with ModelRunner(
         config,
         model,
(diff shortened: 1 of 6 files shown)


bittremieux (commented):
I think this code is way too complex for what it adds. What is the real benefit of having this vs just the total runtime?

Lilferrit:
The total runtime would also include the time it takes to read the data-files and create the data-loaders, which I am not sure is something we necessarily want to include when calculating the spectra sequenced per second. However, this code could definitely be made less complex by simply saving the sequence start and end time (probably in the `ModelRunner`) or by using `lightning.pytorch.callbacks.ThroughputMonitor`, though with the ThroughputMonitor the ThroughPut information would not show up in the end of run report.

bittremieux:
In my opinion there's no need for this functionality. Users only care about the full runtime, it's not relevant whether that's reading the data or doing inference. And if we really need this for timing purposes, we can just as easily extract it from the timestamps that are already logged. And the current implementation is especially not nice, with passing unnecessary information through various levels of the code. So imo this PR should just be closed.

PR #432, opened 2025-02-10
Windows test pipeline failure on `dev` branch fix

Lilferrit:
Fixed windows test pipeline failure on `dev` branch.


Lilferrit:
Turns out the fix for the windows test issue is to explicitly set the weights_only argument in torch.load, since the default changed between pytorch versions. The windows tests pass now.

bittremieux (approved):

PR #434, opened 2025-02-22
Output root suffix fix

Lilferrit:
Fixed issue where the suffix of the output root argument would get overridden, e.g. if output root is set to foo.bar the output files would become `foo.log` and `foo.mztab` rather than the expected `foo.bar.log` and `foo.bar.mztab`.


wsnoble (approved):
Yes, this looks good.

bittremieux (commented):
This works, although I have a suggestion to make it more concise.

Please also list this change in the changelog (after the item discussing splitting `output`).

PR #443, opened 2025-03-17
1st version of beam_search_optimization

NameLessEG:
Reduce the calls to .item() to avoid excessive synchronization between the GPU and CPU. Instead of calling .item() in loops, process tensors in batches. For boolean checks, these should be checked directly on the GPU; if that is not possible, transfer an entire batch to the CPU once with .cpu() and then iterate over it.

Initialize necessary data structures for invalid beam checks in __init__ to prevent redundant setup during runtime.

Optimize detokenization by deferring it until final output in _finish_beams or _cache_finished_beams. Reduce redundant mass calculations by maintaining a beam_masses tensor, updating it incrementally with each token selection rather than recomputing the full sequence mass each time. Perform precursor mass checks using vectorized GPU operations instead of Python loops.

Speed up mass error calculations by leveraging Numba JIT on the CPU if data is already on the CPU. If feasible, implement the logic directly on the GPU or process data in batches on the CPU to minimize GPU-CPU synchronization overhead.

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -5,6 +5,8 @@ import heapq
 import logging
 import warnings
 from typing import Any, Dict, Iterable, List, Optional, Tuple, Union
+import warnings
+warnings.filterwarnings("ignore", category=DeprecationWarning)
 
 import depthcharge.masses
 import einops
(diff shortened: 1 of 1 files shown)


bittremieux:
The overview of the optimizations sounds very interesting, thanks.

Before we proceed with a full code review, can you already address the following:
- Rebase the PR against the `dev` branch, which includes the latest version of the code while we're gearing up for a new release.
- Ensure to our code style by running black.
- The testing code at the end should be removed from the `model.py` file itself.

Based on the timing results, how much do these changes impact the runtime? Very exciting. 🙂

PR #446, opened 2025-03-20
latest version of optimized beam_search_func

NameLessEG:
make sure we can pass all the test cases


casanovo/denovo/model.py
```diff
@@ -528,43 +687,60 @@ def _cache_finished_beams(
...
-                aa_scores, beam_fits_precursor[i]
+                aa_scores, beam_fits_precursor[i].item()
             )
+
             # Omit the stop token from the amino acid-level scores.
             aa_scores = aa_scores[:-1]
-            # Add the prediction to the cache (minimum priority queue,
-            # maximum the number of beams elements).
+
+            # Add the prediction to the cache
```
bittremieux, line 738: **nitpick:** Keep original comment.

tests/unit_tests/test_unit.py
```diff
@@ -13,7 +13,7 @@
 import tempfile
 import unittest
 import unittest.mock
-
+import contextlib
```
bittremieux, line 16: **nitpick:** Sort imports alphabetically within each block.

tests/unit_tests/test_unit.py
```diff
@@ -27,7 +27,12 @@
 from casanovo.data import db_utils, ms_io
 from casanovo.data.datasets import SpectrumDataset, AnnotatedSpectrumDataset
 from casanovo.denovo.evaluate import aa_match, aa_match_batch, aa_match_metrics
-from casanovo.denovo.model import Spec2Pep, _aa_pep_score, _calc_match_score
+from casanovo.denovo.model import (
+    Spec2Pep,
```
bittremieux, line 31: **nitpick:** Sort alphabetically.

tests/unit_tests/test_unit.py
```diff
@@ -1771,3 +1776,218 @@ def test_setup_output(tmp_path, monkeypatch):
...
+        assert mass_sum < expected_mass * 2
+
+
+def test_cache_finished_beams_branches():
+    # Create a model for a single spectrum with 4 beams.
+    model = Spec2Pep(n_beams=4, top_match=2)
+    stop_idx = model.decoder._aa2idx["$"]  # Get the index for the stop token
+
+    # Construct tokens with shape (4, 3) representing 4 beams, each of length 3 (using tokens[:, :3] at step=2)
+    tokens = torch.tensor(
```
bittremieux, line 1829: **nitpick:** Comments can be interleaved different to have the arrays on a single line instead of spread over three lines.

tests/unit_tests/test_unit.py
bittremieux, line 1: **praise:** Great job adding relevant unit tests. 👏

bittremieux (changes requested):
Great work, and the speed-up gains look very impressive. We want to try to get this incorporated for the next Casanovo release.

I've left a few small questions and some suggestions. Based on my understanding of the code, the optimizations should replace the old behavior. In that case I think we can simplify the code by removing the old logic, and making sure that the optimized implementation is always used.

Can you also rebase to the latest version of `dev`. This will require a few small updates, as there were some changes.

You can also remove the "Optimization" word in the comments. That's relevant for this PR, but for future reading after it's merged it doesn't really matter anymore.

Thanks for your contributions!

Issue #402, opened 2024-11-18
Casanovo-DB potential performance improvements


bittremieux:
- Should give a warning when non-supported PTMs are used (e.g. C carbamidomethylation is not specified as a fixed modification) [easy]
- Multithreading during FASTA digestion and m/z calculation [easy]
- Refactor candidate selection to not search from scratch for every candidate but use some sort of looping index [moderate (within batches) — hard (across batches, would require modifications to the data loader)]
- mzTab export should report protein database information [easy]
- Spectrum progress bar should show more granular updates [moderate]
- Superfluous predicted peptide m/z calculation (can be derived from database) [moderate]

Open-ended evaluations:

- Profiling to understand where the runtime and memory consumption bottlenecks are—likely PSM batch creation contributes. Candidate retrieval can probably be optimized using a sliding window approach.
- Investigate whether `_calc_match_score` can be harmonized between de novo and DB modes.

bittremieux:
Fixed in #605.

PR #447, opened 2025-03-24
Add a warning when non-supported PTMs are used.

mikelly11:
This adds warnings for the user if a PTM is not one of the supported PTMS and is added to the config.yaml file. I believe this is mentioned as one of the points in #402.

### DIFF
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -143,6 +143,13 @@ class Config:
         try:
             param_val = self._user_config.get(param, self._params[param])
             if param == "residues":
+                for aa, mass in param_val.items():
+                    aa_str = str(aa)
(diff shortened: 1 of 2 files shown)


wsnoble:
Thank you, but I think this would need to be done in a more general way.  Rather than hard-coding a specific set of modifications (` ["M+15.995", "N+0.984", "Q+0.984", "C+57.021"]:`), we need to use the actual modifications in the model's alphabet and compare those to the modifications in the config file.

mikelly11:
Thank you for the feedback. I have changed this to be done in a general way that compares the config file to the selected model alphabet.

mikelly11:
I made the mistake of not pulling dev. I have edited this and created a new branch with edits from dev that is passing all tests according to pytest.

PR #449, opened 2025-04-02
Warning if config file contains an AA that is not in model alphabel

mikelly11:
I have edited the dev branch with a warning if a PTM or AA is not supported according to the model alphabet. This currently passes all tests on my local instance with pytest. I tested this with sample data and it gives appropriate warnings.

I believe this is mentioned as one of the points in #402.

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -175,6 +175,11 @@ class Spec2Pep(pl.LightningModule, ModelMixin):
         )
         self.stop_token = self.decoder._aa2idx["$"]
 
+        # Check if config PTM and AAs are present in model alphabet
+        for aa in config.Config().residues:
+            if aa not in residues:
+                warnings.warn(f"Warning: {aa} is not in model alphabet.")
+
         # Logging.
         self.calculate_precision = calculate_precision
         self.n_log = n_log


casanovo/denovo/model.py
```diff
@@ -175,6 +175,11 @@ def __init__(
         )
         self.stop_token = self.decoder._aa2idx["$"]
 
+        # Check if config PTM and AAs are present in model alphabet
+        for aa in config.Config().residues:
```
bittremieux, line 179: **todo:** `config.Config()` creates a default config, whereas the active config might be loaded from a different file.

Instead, it's probably better to do this one level up, in `model_runner.py`, where the active config is used. Conveniently, PR #442 introduces a `initialize_tokenizer()` method which seems a good location to add this warning. So you can rebase to that one and then add the functionality there.

Can you also add some unit tests that check that the warning is correctly reported?

casanovo/denovo/model.py
```diff
@@ -175,6 +175,11 @@ def __init__(
...
+            if aa not in residues:
+                warnings.warn(f"Warning: {aa} is not in model alphabet.")
```
bittremieux, line 181: **nitpick:** Maybe slightly update the message to also mention that `aa` is specified in the config, so that it's clear to the user where the discrepancy comes from. The "Warning" is also not really necessary, because Python itself will say that it's a warning.

Something like this?

> Configured {aa} residue is not in model alphabet.

bittremieux (changes requested):

bittremieux:
Update: #442 has been merged into `dev`, so please rebase against the latest version.

JosieHong:
> @JosieHong Considering that you were making some config-related improvements, do you mind also taking a stab at this? It's probably easiest to start a new PR with the changes as suggested in my review.

Not at all! I’ll open a new PR with the changes you suggested.

bittremieux:
Superseded by #538.

PR #456, opened 2025-04-20
Log system information at start

### LINKED ISSUE #422: Log certain run information at the start of a run instead of the end.
Currently a lot of the sequencing/training run information is logged at the very end of a run, but a lot of this information such as the machine hostname and full command can be useful for debugging purposes if a run is terminated early due to an error. It might be worthwhile to log some of these entries at the very beginning of the run instead of the end.

bittremieux:
Fixes #422.


casanovo/utils.py
```diff
@@ -151,24 +105,54 @@ def get_report_dict(
...
 
-    peptide_lengths = get_peptide_lengths(results_table["sequence"])
-    min_length, med_length, max_length = np.quantile(
-        peptide_lengths, [0, 0.5, 1]
+    # Mass modifications do not contribute to sequence length.
+    # FIXME: If PTMs are represented in ProForma notation this filtering
+    #  operation needs to be reimplemented.
+    pep_lens = (
+        results_table["sequence"]
+        .str.replace(r"[^a-zA-Z]", "", regex=True)
```
Lilferrit, line 113: Using the split method of the depthcharge PeptideTokenizer might be a more straight forward way of doing this, and would probably make it easier to filter out mass modifications and support the ProForma format.

Lilferrit (approved):

casanovo/utils.py
bittremieux: Yes, you're right. There are a few spots in the code with some remnant peptide parsing that needs to be fixed. This is pre-existing code not directly impacted by this PR though, so I'll merge this first..

PR #458, opened 2025-04-21
Switched peptide level score to product of AA scores and removed AA score adjustement step

Lilferrit:


casanovo/denovo/model.py
```diff
@@ -1436,8 +1435,7 @@ def _aa_pep_score(
     peptide_score : float
         The peptide score.
     """
-    peptide_score = np.exp(np.mean(np.log(aa_scores)))
-    aa_scores = (aa_scores + peptide_score) / 2
```
Lilferrit, line 1440: **question**: Since the AA score adjustment step has been removed should this function maybe be changed to not return the AA scores anymore? I kept this as is to keep the initial PR as minimal as possible.

bittremieux (commented):
Simplifying the method to only return the peptide score makes sense.

Also make sure to modify the entry in the changelog (overwriting the previous one about geometric mean as well as adding a new entry for the amino acid scores).

Issue #461, opened 2025-04-22
Depthcharge reverse detokinization bug causes inference to fail in some cases.


Lilferrit:
In particular, the incorrect behavior behavior when detokenizing reversed sequences causes the sequences to be in an incorrect format, which then causes this block in `on_predict_batch_end` to throw an error:

```python
# Compute the precursor m/z of the predicted peptide.
spec_match.calc_mz = self.tokenizer.calculate_precursor_ions(
    spec_match.sequence, torch.tensor(spec_match.charge)
).item()
```

See: https://github.com/wfondrie/depthcharge/pull/72

PR #462, opened 2025-04-22
Depthcharge reverse detokinization bug work-around

Lilferrit:
Workaround to fix issue described #461


bittremieux (approved):
Keep the original code in there but commented out and marked with a `#FIXME` so that it's easier to revert this workaround as soon as the bug is fixed within Depthcharge.

PR #463, opened 2025-04-22
Add ground truth sequences, peptide correct call, and amino acid correct calls to MzTab when running casanovo in evaluate mode

### LINKED ISSUE #453: improve evaluate outputs
Currently, the evaluate option simply adds three lines to the log file, like this:

`2025-04-11 17:32:38,752 INFO [casanovo/MainProcess] model_runner.log_metrics : Peptide Precision: 49.23%
2025-04-11 17:32:38,753 INFO [casanovo/MainProcess] model_runner.log_metrics : Amino Acid Precision: 71.18%
2025-04-11 17:32:38,753 INFO [casanovo/MainProcess] model_runner.log_metrics : Amino Acid Recall: 71.28%`

I think we should also report average precision, and we should output a separate TSV file that allows for plotting of the precision-coverage curve.  This could include the Casanovo score, ground truth and predicted peptides, correct/incorrect label, and the precision.

Lilferrit:


bittremieux (commented):
Just a few quick thoughts before a full review, as the PR is still marked as in-progress.
- I don't think we need an extra `AnnotatedPepSpecMatch` class, you can just add an optional `sequence_true` attribute to the existing `PepSpecMatch` class.
- You're now doing a lot of extra stuff in `model_runner.py`, which doesn't seem the appropriate place for that.
- Avoid mixing in stuff from other PRs here as well, as it makes it a bit trickier to understand what has changed. If you need those changes, open this PR on top of the other one rather than `dev`.

PR #469, opened 2025-05-06
Fix multi GPU training

Lilferrit:


Lilferrit:
I added an option that if enabled will force PyLightning to use IPv4 for inter-process communication in the multi-GPU settings, because some environments (such as our own cluster) don't support IPv6.

casanovo/denovo/model_runner.py
```diff
@@ -351,7 +351,10 @@ def initialize_trainer(self, train: bool) -> None:
                     tb_log_dir = "tensorboard"
 
                     if self.config.log_metrics:
-                        if self.overwrite_ckpt_check:
+                        if (
```
Lilferrit, line 354: Only the main process will write checkpoints and logs, so only the main process needs to do the overwrite check.

casanovo/casanovo.py
```diff
@@ -637,6 +637,9 @@ def _setup_output(
             f"casanovo_{datetime.datetime.now().strftime('%Y%m%d%H%M%S')}"
         )
 
+    if utils.get_local_rank() != 0:
```
Lilferrit, line 640: The tricky thing here is if an output_root isn't provided each of the sub-process will have a different root-name since they will be started at different times. Unfortunately I don't think this is really solvable without setting additional environmental variables in the main process.

Lilferrit: A potential solution to this could look like this, but I'm not sure if the additional complexity would be worth it:

```python
def _setup_output(
    output_dir: str | None,
    output_root: str | None,
    overwrite: bool,
    verbosity: str,
) -> Tuple[Path, str]:
    """
    Set up the output directory, output file root name, and logging.

    Parameters:
    -----------
    output_dir : str | None
        The path to the output directory. If `None`, the output
        directory will be resolved to the current working directory.
    output_root : str | None
        The base name for the output files. If `None` the output root
        name will be resolved to casanovo_<current date and time>
    overwrite: bool
        Whether to overwrite log file if it already exists in the output
        directory.
    verbosity : str
        The verbosity level for logging.

    Returns:
    --------
    Tuple[Path, str]
        A tuple containing the resolved output directory and root name
        for output files.
    """
    output_root = os.getenv("CASA_OUTPUT_ROOT", output_root)
    if output_root is None:
        output_root = (
            f"casanovo_{datetime.datetime.now().strftime('%Y%m%d%H%M%S')}"
        )
        os.environ["CASA_OUTPUT_ROOT"] = output_root
    ...
```

bittremieux (commented):
**question:** So when doing inference this will produce multiple mzTab output files, one for each GPU? This is not super user-friendly imo, because then the user won't know where specific results can be found.

I'm also not sure what exactly the problem is.
1. Multi-GPU training should work correctly, or at least it has for a long time.
2. Multi-GPU inference indeed doesn't properly work. That's also why we [set the number of devices to 1](https://github.com/Noble-Lab/casanovo/blob/main/casanovo/denovo/model_runner.py#L329) and only increase it during training mode. (This is also not changed in this PR, so multi-GPU inference still wouldn't happen I think.) But doing inference can be trivially parallelized by executing separate commands on different files, so I don't think that this is a major issue either. At least to the extent that any solution should clearly improve upon this status quo.

Lilferrit:
> Multi-GPU training should work correctly, or at least it has for a long time.

Mutli-GPU training was broken by the changes to the output file name/locations in #372. Since each sub-process will create it's own log file, if an `output_root` is provided after the main process creates a log file with this name subsequent sub-processes will attempt to create a log file with the same name which will lead to either that process crashing or that log file being silently overwritten. I felt the best way to deal with this was just to divert each sub-process's logging to a separate log file.

> Multi-GPU inference indeed doesn't properly work

Yep correct. I initially drafted this PR several months ago before revisiting it so I misremembered what exactly the scope was.

bittremieux (changes requested):

PR #478, opened 2025-05-20
Casanovo DB Peptide Sequence Fix

Lilferrit:


bittremieux (commented):
The implementation is fine. But as mentioned on Slack, maybe this is not the ideal solution. While in a de novo context equating leucine and isoleucine makes sense, there's not really a reason to do this during database searching.

So instead of these workarounds, maybe we should just ignore this setting in DB mode and always set it to false?

bittremieux:
That's true. Can you add a quick unit test that verifies this behavior before merging?

PR #480, opened 2025-06-13
Casanovo DB AA Score Fix

Lilferrit:
Marking this PR as a draft pending further testing as well as updating `test_calc_match_score`.


Lilferrit:
Not sure what's happening with the failed test because I do not see the same behavior on my machine.

Lilferrit:
denovo_v_db_fixed
It looks like the scores match the de novo scores now.

Lilferrit:
I also added a reverse amino acid score fix for Casanovo DB to this PR.

casanovo/denovo/model.py
```diff
@@ -1339,32 +1339,24 @@ def _calc_match_score(
...
+    # Get aa scores corresponding with true aas
+    true_aa_scores = torch.gather(
+        batch_all_aa_scores, 2, truth_aa_indices.unsqueeze(-1)
+    ).squeeze(-1)
+
+    # Calculate peptide scores and aa scores
+    pep_scores_all, aa_scores_all = [], []
+    aa_score_mask = truth_aa_indices != 0
+    for aa_scores, aa_score_mask in zip(true_aa_scores, aa_score_mask):
+        aa_scores = aa_scores[aa_score_mask].cpu().detach().numpy()
```
bittremieux, line 1354: *question:** The previous approach, moving the entire `aa_scores` and `aa_score_mask` tensors to the CPU in one go, rather than within the loop, is probably more efficient?

I don't think this part of the code needed to be changed? The functionality seems the same, except for renamed variables.

bittremieux (changes requested):

casanovo/denovo/model.py
Lilferrit: Yep I just got a bit overzealous while rewriting this section, I reverted to the old implementation for this section of the code.

PR #481, opened 2025-06-17
Casanovo DB Add Stop Token

Lilferrit:


Lilferrit:
Marking this as a draft for now since this depends on a fix in #480.

bittremieux (commented):
This looks good. Can you include a unit test that verifies that stop tokens are properly handled?

bittremieux:
The `test_get_candidates_isotope_errror` test seems to fail.

bittremieux (approved):

PR #482, opened 2025-06-17
Add Option to Disable Precursor Check

Lilferrit:
Marking as draft since Github tests are failing pending #480.

### DIFF
--- a/casanovo/config.yaml
+++ b/casanovo/config.yaml
@@ -14,6 +14,7 @@
 # peptide score.
 # database search: Select candidate peptides within the specified precursor m/z
 # tolerance.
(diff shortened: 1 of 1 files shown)


bittremieux (commented):
While functional. I don't think that this is the best solution. If no precursor mass tolerance is set, we can skip several calculations, rather than using an infinite tolerance as a workaround.

bittremieux:
Yes, it should build on top of #470. I think tying it into the decoding strategy makes sense here. While it will be a bit more involved, the code changes should still be minimal, and it would also make the decoding logic more efficient in that case.

PR #483, opened 2025-06-18
Stable product score

Lilferrit:


casanovo/denovo/model.py
```diff
@@ -1453,7 +1453,9 @@ def _peptide_score(aa_scores: np.ndarray, fits_precursor_mz: bool) -> float:
     peptide_score : float
         The peptide score.
     """
-    peptide_score = np.prod(aa_scores)
+    tol = 10**-13
+    aa_scores[aa_scores < tol] = tol
```
bittremieux, line 1457: **todo:** Maybe something like this would be a better solution than the `10**-13` "magic number":
```
aa_scores = np.clip(aa_scores, np.finfo(np.float64).eps, 1)
```

PR #489, opened 2025-07-01
Casanovo DB Merge N-terminal Scores with Leading AA

Lilferrit:


tests/unit_tests/test_unit.py
```diff
@@ -1774,9 +1821,11 @@ def test_beam_search_decode(tiny_config):
...
     try:
         # Test _finish_beams()
-        finished_beams, beam_fits_precursor, discarded_beams = (
-            model._finish_beams(tokens, precursors, step)
-        )
+        (
+            finished_beams,
+            beam_fits_precursor,
+            discarded_beams,
+        ) = model._finish_beams(tokens, precursors, step)
```
Lilferrit, line 1828: Not sure why the pre-commit hook changed these.

casanovo/denovo/model.py
```diff
@@ -1409,6 +1409,14 @@ def predict_step(
                 spectrum_id = (filename, scan)
                 if self.tokenizer.reverse:
                     curr_aa_scores = curr_aa_scores[::-1]
+
+                # Merge n-terminal mod score with leading AA if applicable
```
bittremieux, line 1413: **todo:** This now only does it for database searching, whereas it should similarly be done for de novo results. Ideally without too much code duplication, so abstract this at a higher level.

tests/unit_tests/test_unit.py
bittremieux: Sometimes it happens between different versions of black.

bittremieux (changes requested):
The key issue is that N-terminal AA scores need to be accounted for in any mode, so for DB searching but also for de novo.

Also please reflect this in the CHANGELOG then.

bittremieux (approved):
I forgot that this had already been added to the base class. LGTM.

PR #504, opened 2025-08-10
DB-search speed up

NameLessEG:
1. Run the encoder once per spectrum batch; score candidates with decoder-only.
2. Generate candidate batches on-device.Less copying.
3. Do encoding outside the candidate loop in predict_step; forward supports cache passthrough for fast scoring.
4. 2.1x speed up

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -5,7 +5,7 @@ import heapq
 import itertools
 import logging
 import warnings
-from typing import Any, Dict, Generator, Iterable, List, Optional, Tuple, Union
+from typing import Any, Dict, Iterable, List, Optional, Tuple
 
 import einops
 import lightning.pytorch as pl
@@ -866,8 +866,7 @@ class Spec2Pep(pl.LightningModule):
         ).float()
         # Mask out the index '0', i.e. padding token, by default.
         active_mask[:, :beam] = 1e-8
-
-        # Ensure we use the exact same scoring and topk mechanism as original code
+        # Compute beam scores and select top-k candidates
         # Use nanmean to properly handle NaN values in scores
         mean_scores = torch.nanmean(step_scores, dim=1)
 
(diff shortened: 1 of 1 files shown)


casanovo/denovo/model.py
```diff
@@ -1378,78 +1405,94 @@ def predict_step(
...
 
-                spectrum_id = (filename, scan)
-                if self.tokenizer.reverse:
-                    curr_aa_scores = curr_aa_scores[::-1]
+        with torch.inference_mode():
+            # Pre-compute encoder outputs for the entire batch.
+            memories, mem_masks = self.encoder(
+                batch["mz_array"], batch["intensity_array"]
+            )
+            precursor_mzs = batch["precursor_mz"].squeeze(0).to(self.device)
```
bittremieux, line 1418: **question:** Can we reuse `_process_batch` for this? The code is almost entirely the same.

casanovo/denovo/model.py
```diff
@@ -1378,78 +1405,94 @@ def predict_step(
...
             pred.protein = self.protein_database.get_associated_protein(
                 pred.sequence
             )
 
         return predictions
 
-    def _psm_batches(
-        self, batch: Dict[str, torch.Tensor]
-    ) -> Generator[Dict[str, torch.Tensor], None, None]:
+    def _psm_batches(self, batch, enc_cache=None):
```
bittremieux, line 1495: **todo:** Add type hints back.

casanovo/denovo/model.py
```diff
@@ -1473,60 +1516,95 @@ def _psm_batches(
...
+        # Iterate precursor charges and m/z values per spectrum.
+        charge_iter = batch["precursor_charge"]  # tensor[B]
+        mz_iter = batch["precursor_mz"]  # tensor[B]
+
+        # Use pre-computed encoder outputs if available; otherwise compute once here.
+        if enc_cache is None:
+            memories, mem_masks = self.encoder(
+                batch["mz_array"], batch["intensity_array"]
+            )
+            precursor_mzs = batch["precursor_mz"].squeeze(0).to(device)
```
bittremieux, line 1542: **suggestion:** Here again an opportunity to reuse `_process_batch`.

casanovo/denovo/model.py
```diff
@@ -1567,15 +1645,30 @@ def _calc_match_score(
...
+
+    # Fuse scores and lengths for a single GPU->CPU transfer
+    fused = torch.cat(
+        [per_aa_scores, lengths.to(per_aa_scores.dtype).unsqueeze(1)], dim=1
+    )
+    fused_np = fused.detach().cpu().numpy()
+
+    # Unpack scores and lengths on the CPU
+    per_aa_np = fused_np[:, :-1]
+    lengths_np = fused_np[:, -1].astype(np.int64, copy=False)
```
bittremieux, line 1659: **nitpick:** Probably `int32` should be sufficient? Or even `int8` actually.

casanovo/denovo/model.py
```diff
@@ -1567,15 +1645,30 @@ def _calc_match_score(
...
     # Calculate peptide scores and aa scores
-    per_aa_scores = per_aa_scores.cpu().detach().numpy()
-    score_mask = (truth_aa_indices != 0).cpu().detach().numpy()
-    peptide_scores, aa_scores = [], []
-    for psm_score, psm_mask in zip(per_aa_scores, score_mask):
-        psm_aa_scores = psm_score[psm_mask]
-        psm_peptide_score = _peptide_score(psm_aa_scores, True)
-        peptide_scores.append(psm_peptide_score)
-        aa_scores.append(psm_aa_scores)
+    eps = np.finfo(np.float64).eps
```
bittremieux, line 1662: **todo:** We should reuse `_peptide_score` so that there's only one code location that does this and that's reused.

NameLessEG: So The main reason I wrote a separate piece of code in _calc_match_score was for performance.
it processes the entire batch of peptides using numpy, which is much faster. The original _peptide_score function is set up to handle only a single peptide at a time.If I reuse _peptide_score，I have to add a for loop, and after experiement, there is a 33.33% increase in execution time.

bittremieux: Yes, the efficiency gain makes sense. However, I'd still strongly like to avoid duplicating peptide score calculations. We've changed how this score is calculated in the latest release, and doing that is only easy to do if there's a single location in the code where it happens.

The batched calculation of the peptide score is good, we want to do it that way. But the calculation should still be harmonized between both the de novo mode and the database search mode.
1. The de novo mode should also start using batched peptide score calculation. Maybe there are some performance improvements there still as well?
2. Or the method should transparently support calculating the score for both a single PSM and for batched PSMs. It seems like this should certainly be feasible.

PR #519, opened 2025-09-22
Fixed coverage, linting, and loading issues on previous pull request

TheBobBob:
Implemented output-db feature for debugging, and wrote tests for no model for db search, peptide digestion with unknown amino acids, and reverse predictions for peptides.


casanovo/casanovo.py
```diff
@@ -262,6 +270,11 @@ def db_search(
 
         results_path = output_path / f"{output_root_name}.mztab"
         runner.db_search(peak_path, fasta_path, str(results_path))
+        if output_db:
+            logger.info(
+                "Completed database search, moving onto dumping for debugging"
+            )
+            runner.model.protein_database.output_db(output_db)
```
Lilferrit, line 277: **Praise**: I really like how this is implemented as an addition to the `protein_database` interface - very clean.

tests/unit_tests/test_unit.py
```diff
@@ -37,6 +46,335 @@
...
+    assert all(
+        [
+            np.allclose(np.array([0.0, 0.0, 0.0]), x)
+            for x in aa_scores_reversed[:4]
+        ]
+    )
+    assert np.allclose(np.array([0.0]), aa_scores_reversed[4])
+
+
+def test_output_db(tiny_fasta_file, mgf_small, tiny_config_db):
```
Lilferrit, line 147: Here it would be better to test from the level of the `ProteinDatabase`. In this case that would probably look like:

1. Initializing the protein database using a small fasta file, maybe `tiny_fasta_file`
2. Checking the output file for correctness.

If the database output is implemented similarly to the one liner I suggested above step 2 could be done by just loading in the database dump as a pandas dataframe and using:

```python
assert loaded_df = protein_db.db_peptides
```

Lilferrit (changes requested):

tests/unit_tests/test_unit.py
Lilferrit: Ah I see. In that case we can probably get around this by still doing this specific test at the level of `ProteinDatabase`, but in the integration test add the `--output_db` argument for Casanovo DB (also probably assert that the db file is written). That way the output_db branch in the CLI is covered by the integration test and hopefully the codecov bot will be happy.

bittremieux: Note: Unit tests should _never_ depend on getting model weights from GitHub.

TheBobBob: Thank you! I added the output_db test to the integration tests, and ensured that none of my tests rely on model weights. I will make a pull request later today with my changes.

Issue #285, opened 2024-01-29
Rank of updates array (1) must be greater than or equal to inner-most dimension of indices array (29)


wsnoble:
On an M2 Mac, I am getting the error shown below.  I do have the fallback variable set properly:

```
% echo $PYTORCH_ENABLE_MPS_FALLBACK
1
```

I am using the following version:

```
Casanovo: 4.0.1
Depthcharge: 0.2.3
Lightning: 2.1.3
PyTorch: 2.1.2
```

The input MGF file is [151009_exo4_4.short.mgf.txt](https://github.com/Noble-Lab/casanovo/files/14086675/151009_exo4_4.short.mgf.txt)

```
% casanovo sequence 151009_exo4_4.short.mgf 
WARNING: Dataloader multiprocessing is currently not supported on Windows or MacOS; using only a single thread.
Seed set to 454
INFO: Model weights file /Users/wnoble/Library/Caches/casanovo/casanovo_massivekb_v4_0_0.ckpt retrieved from local cache
INFO: Casanovo version 4.0.1
INFO: Sequencing peptides from:
INFO:   151009_exo4_4.short.mgf
GPU available: True (mps), used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
INFO: Reading 1 files...
151009_exo4_4.short.mgf: 2000spectra [00:00, 20073.48spectra/s]
WARNING: Skipped 1282 spectra with invalid precursor info
Predicting DataLoader 0:   0%|                            | 0/1 [00:00<?, ?it/s]WARNING: UserWarning: The operator 'aten::_nested_tensor_from_mask_left_aligned' is not currently supported on the MPS backend and will fall back to run on the CPU. This may have performance implications. (Triggered internally at /Users/runner/work/pytorch/pytorch/pytorch/aten/src/ATen/mps/MPSFallback.mm:13.)
WARNING: UserWarning: The operator 'aten::_nested_tensor_from_mask_left_aligned' is not currently supported on the MPS backend and will fall back to run on the CPU. This may have performance implications. (Triggered internally at /Users/runner/work/pytorch/pytorch/pytorch/aten/src/ATen/mps/MPSFallback.mm:13.)
[...]

PR #522, opened 2025-09-23
Set trainer's device to CPU if it is initially loaded onto MPS

Lilferrit:


casanovo/denovo/model_runner.py
```diff
@@ -400,6 +400,14 @@ def initialize_trainer(self, train: bool) -> None:
 
         self.trainer = pl.Trainer(**trainer_cfg)
 
+        # Reinitialize the trainer if it was loaded on to mps
+        if self.trainer.strategy.root_device.type == "mps":
```
bittremieux, line 404: **question:** Is the device available before the trainer is initialized? If so, I'd keep changing from MPS to CPU consistent with how the number of workers is set (i.e. utils function and modify it [directly during config loading](https://github.com/Noble-Lab/casanovo/blob/ec67e980019a721ec91debe1068692ba78e13ac3/casanovo/config.py#L154). That has as benefit that config settings are interpreted in a single location, and makes it easier to add more logic if other accelerators also have compatibility issues. (+ Small extra advantage because of not having to initialize the trainer twice.)

Lilferrit: This was unsatisfying to me too but alas it doesn't look like there is really a way to tell how pytorch lightning will resolve `accelerator="auto"` without initializing the trainer first, nor is there a way to move the trainer once it's initialized.

Another way to do this would be to just check if the machine is Apple silicon and then setting the accelerator to CPU if so, which could be done in the config object, however I feel like this would introduce an edge case if a user is on Apple Silicon but also has a GPU available.

Lilferrit:
Sure thing, I just updated the changelog.

Lilferrit:
I just tested the MGF file attatched to #285 on an m.3 Macbook and this fix does resolve the issue when using the default config.

PR #523, opened 2025-09-23
Strip modifications and add modifcations to correct column in the output MzTab file as per the MzTab specification.

Lilferrit:


casanovo/data/ms_io.py
```diff
@@ -144,6 +145,87 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
             )
             self._run_map[Path(filename).name] = i
 
+    @staticmethod
+    def get_mod_string(
+        mod: pyteomics.proforma.ModificationBase,
+        residue: str = "N-term",
```
bittremieux, line 151: **todo:** It might be better to have `None` as the default value. Or rather no default value, so that the caller needs to specify it. I don't think there's anything special about N-term / position 0.

casanovo/data/ms_io.py
```diff
@@ -144,6 +145,87 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+        position: int = 0,
```
bittremieux, line 152: **todo:** Same here.

casanovo/data/ms_io.py
```diff
@@ -144,6 +145,87 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+            N-terminal modifications, use `"N-term"`.
+        position : int, default=0
+            Position of the modification in the peptide sequence.
+            Use `0` for N-terminal, `len(sequence)+1` for C-terminal,
+            and a 1-based index for internal residues.
+
+        Returns
+        -------
+        str
+            The mzTab-formatted modification string. Examples:
```
bittremieux, line 174: **todo:** No examples actually given.

casanovo/data/ms_io.py
```diff
@@ -144,6 +145,87 @@ def set_ms_run(self, peak_filenames: List[str]) -> None:
...
+
+        Returns
+        -------
+        aa_seq : str
+            The plain amino acid sequence with modifications stripped.
+        mod_string : str
+            A semicolon-delimited string of modifications in mzTab
+            format, suitable for reporting in the PSM section.
+        """
+        seq_mod, term = pyteomics.proforma.parse(seq)
```
bittremieux, line 205: **todo:** The necessary information might be slightly more directly available through spectrum_utils' ProForma functionality:
- https://spectrum-utils.readthedocs.io/en/latest/api.html#spectrum_utils.proforma.Proteoform
- https://spectrum-utils.readthedocs.io/en/latest/api.html#spectrum_utils.proforma.Modification

casanovo/data/ms_io.py
bittremieux, line 299: **todo:** Add an optional column `op_ms_run[1]_proforma` that contains the full ProForma string, for backwards compatibility.

bittremieux (changes requested):
Looks good, but maybe just a few small tweaks.

Also add a note to the changelog.

bittremieux:
#525 can build on top of this, so maybe you could tackle it in one go.

casanovo/utils.py
```diff
@@ -106,11 +106,9 @@ def _get_report_dict(
         return None
 
     # Mass modifications do not contribute to sequence length.
-    # FIXME: If PTMs are represented in ProForma notation this filtering
-    #  operation needs to be reimplemented.
     pep_lens = (
         results_table["sequence"]
-        .str.replace(r"[^a-zA-Z]", "", regex=True)
+        .str.replace(r"\[.*?\]-?", "", regex=True)
```
Lilferrit, line 111: My initial thought was to add the parsed AA sequence to the `PeptideSpectrumMatch` data-classes as they are parsed in the output writer, but unfortunately this doesn't happen until the context manager for the `ModelRunner` exits, which happens as the run report is logged.

casanovo/data/psm.py
```diff
@@ -1,7 +1,7 @@
 """Peptide spectrum match dataclass."""
 
 import dataclasses
-from typing import Iterable, Tuple
+from typing import Iterable, Optional, Tuple
```
bittremieux, line 4: **nitpick:** No longer needed?

Lilferrit:
Good idea, I'll go ahead and implement this way instead.

casanovo/data/psm.py
```diff
@@ -43,3 +45,74 @@ class PepSpecMatch:
...
+            residue = aa_seq[mod.position]
+
+        for src in mod.source or []:
+            if hasattr(src, "accession"):
+                return f"{pos}-{src.name} ({residue}):{src.accession}"
+
+        return f"{pos}-[{mod.mass:+.4f}]"
+
+    @property
+    def _proteoform(self) -> spectrum_utils.proforma.Proteoform:
```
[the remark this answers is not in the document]
Lilferrit: I did initially implement this via caching the ProForma parsing results, but then I realized if the `seqeunce` field gets changes the cached parsing results would no longer be valid.

A way around this would be to mark the dataclass as immutable using the `frozen` argument of the class decorator - since none of the fields are references to mutable objects and AFIK nowhere in the code actually modifies the fields after initialization. Using the `frozen` flag would just ensure that doing so in the future would raise an exception.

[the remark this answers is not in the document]
Lilferrit: Alternatively we could also keep a reference pointing the the `sequence` string at the time the cache is initialized, and re-parse if it changes upon a cache access, but imo that's a starting to get bit over-complicated for a dataclass.

PR #529, opened 2025-10-03
PTMs don't count towards peptide length

NiklasFeurstein:
The calculation of peptide length did not work for the ProForma notation.


bittremieux:
Thanks for the contribution @NiklasFeurstein. Coincidentally we were just addressing this in #523 as well, so I'll close this PR.

PR #536, opened 2025-10-14
Changed force overwrite

bhavyak7-lab:
Changed force overwrite to overwrite mztab files as well. 
Changes made:
- moved file overwriting check (if overwrite check is false) to ModelRunner
- removed lines checking for the mztab files from casanovo.py


bittremieux:
Superseded by #612.

PR #538, opened 2025-10-17
Warning if config file contains an AA that is not in model alphabel (follow-up)

josiehong:
* Added a warning in `initialize_tokenizer()` to notify when residues specified in the user configuration are not present in the tokenizer’s residue alphabet.
* Added a corresponding unit test (`test_initialize_tokenizer()`) in `tests/unit_tests/test_runner.py`.
* Updated the warning message to list all missing residues from the configured residue set for better readability.


casanovo/denovo/model_runner.py
```diff
@@ -407,6 +407,14 @@ def initialize_tokenizer(self) -> None:
         else:
             tokenizer_clss = PeptideTokenizer
 
+        missing_aa = list(
+            set(self.config.residues) - set(tokenizer_clss.residues)
+        )
+        if missing_aa:
+            warnings.warn(
```
bittremieux: **todo:** It's probably better to use `logger.warning(...)` here, because this is a user-facing message.

JosieHong:
I’ve made the requested updates. Could you please review them again?

bittremieux (approved):
Looks good. Can you also add a quick note in the changelog before merging?

PR #542, opened 2025-10-23
Check tokenizer equivalence via residue and index dicts rather than memory address

Lilferrit:


casanovo/denovo/model_runner.py
```diff
@@ -522,6 +522,12 @@ def initialize_model(self, train: bool, db_search: bool = False) -> None:
             )
             for param in architecture_params:
                 if model_params[param] != self.model.hparams[param]:
+                    if param == "tokenizer":
+                        self._verify_tokenizer(
+                            model_params[param], self.model.hparams[param]
+                        )
+                        continue
```
bittremieux: **nitpick:** I prefer `if ... else` for more explicit logic.

tests/unit_tests/test_runner.py
```diff
@@ -2,6 +2,7 @@
 
 import shutil
 import unittest.mock
+import warnings
```
bittremieux: **question:** Import still needed?

PR #546, opened 2025-11-01
Evaluation metrics bug fix

Lilferrit:
Fixed bug that led to the reported peptide precision always being 0.


bittremieux (approved):
Great. 👍

Please still update the CHANGELOG before merging.

PR #548, opened 2025-11-03
Updating NonAR to 5.0

devanshidesai25:
Changed the model from autoregressive to non-autoregressive decoding by disabling causal masking, feeding zero tokens to the decoder instead of ground truth tokens during training, enabling all amino acid positions to be predicted simultaneously in parallel. Also removed the [:, :-1] slicing in training_step to predict all positions including the stop token.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -307,7 +307,7 @@ def _get_config_file(file_path, file_name, additional_cfg=None):
         "weight_decay": 1e-5,
         "train_label_smoothing": 0.01,
         "train_batch_size": 32,
-        "max_epochs": 20,
+        "max_epochs": 50,
         "shuffle": False,
         "shuffle_buffer_size": 64,
         "num_sanity_val_steps": 0,


casanovo/denovo/model.py
```diff
@@ -1148,51 +412,118 @@ def validation_step(
...
+            precursors=precursors,
+        )
+
+        # Apply N-terminal modification masking
+        B, L, V = logits.shape
+        nterm_idx = self.nterm_idx.to(device)
+        
+        if self.tokenizer.reverse:
+            # N-term mods only allowed at last position
+            if L > 1:
```
Lilferrit: I don't think the logic here is correct. Since peptides can be different lengths you could have a valid n-terminal modification say in position 5 in one peptide in a batch but position 7 at another peptide in the batch. The logic for the n to c decoding case is correct though.

casanovo_20251103225417.mztab
bittremieux, line 1: **todo:** Must be removed.

casanovo_5_train.sge
bittremieux, line 1: **todo:** Must be removed.

casanovo_sanity_check.sge
bittremieux, line 1: **todo:** Must be removed.

+
bittremieux, line 1: **todo:** Must be removed.

bittremieux (changes requested):

PR #549, opened 2025-11-04
Removed vocabulary warning

Lilferrit:
From my understanding this warning wasn't actually really doing anything. The code that pops the warning just compares the config's vocabulary against depthcharge's default tokenizer vocabulary - which is just the canonical amino acids.

If the checkpoint and config tokenizer vocabularies actually don't match, a warning is already popped by `modelrunner._verify_tokenizer`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 - A mismatching parameter warning will now only be triggered for the tokenizer if the config and checkpoint tokenizers do not have equivalent vocabularies.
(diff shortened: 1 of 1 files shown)


tests/unit_tests/test_runner.py
```diff
@@ -541,21 +541,6 @@ def get_mock_psm(sequence, spectrum_id):
             assert aa_recall == pytest.approx(100 * (2 / 3))
 
 
-def test_initialize_tokenizer(caplog):
```
Lilferrit: This unit test tests the functionality that was removed in this PR, which is also now redundant with `_verify_tokenizer`.

bittremieux (approved):
Looks good. Please still update the CHANGELOG.

PR #550, opened 2025-11-04
Fix tests on Apple Silicon

Lilferrit:


Lilferrit:
The Casanovo DB tests were failing due to floating point rounding differences between platforms. I updated these unit tests to compare the mass values instead of the actual peptide sequences, since I don't think that testing the lexicographical order tie breaker is really of the upmost importance.

PR #555, opened 2025-11-12
Added timsTOF support

bhavyak7-lab:
Changes made
- Fixed linting issue 
- Wrote tests
- Added new file format to file_formats.md
- Updated CHANGELOG.md

## Summary by CodeRabbit

* **New Features**
  * timsTOF files in Bruker .d format are now supported as spectra input for de novo peptide sequencing.

* **Documentation**
  * Updated supported input file formats to reflect Bruker D format compatibility.

* **Tests**
  * Added a unit test validating loading of .d folders and handling of unsupported file extensions.

* **Changelog**
  * Added an "Added" entry noting .d folder support.

### DIFF
--- a/tests/unit_tests/test_runner.py
+++ b/tests/unit_tests/test_runner.py
@@ -37,11 +37,10 @@ def test_loading_timstof_folders(tmp_path, monkeypatch):
     # (should raise FileNotFoundError and RuntimeWarning)
     fake_path = tmp_path / "test.hi"
     fake_path.mkdir()
-    with pytest.warns(RuntimeWarning):
-        with pytest.raises(FileNotFoundError):
-            runner._get_input_paths(
-                peak_path=(str(fake_path),), annotated=False, mode="validation"
-            )
+    with pytest.warns(RuntimeWarning), pytest.raises(FileNotFoundError):
+        runner._get_input_paths(
+            peak_path=(str(fake_path),), annotated=False, mode="validation"
+        )
 
 
 def test_initialize_model(tmp_path, mgf_small):


casanovo/denovo/model_runner.py
```diff
@@ -718,6 +686,10 @@ def _get_peak_filenames(
         for fname in glob.glob(path, recursive=True):
             if Path(fname).suffix.lower() in supported_ext:
                 found_files.add(fname)
+            elif Path(fname).is_dir() and any(
```
bittremieux: **todo:** This is still not correct, because the code will never get here. A ".d" folder will also trigger the suffix check on line 687 and this `elif` will not be accessed.

Your tests still work because even though `.d` is a folder, it needs to be handled as a file. And these added lines of code are not even needed as far as I can see.

casanovo/denovo/model_runner.py
```diff
@@ -718,6 +686,10 @@ def _get_peak_filenames(
...
+                Path(fname).name.lower().endswith(ext) for ext in supported_ext
```
bittremieux: **question:** Why does this need to an an `any` - `endswith` and not simple `in` the same way as above?

Anyway, this file type reasoning needs to be properly revisited.

docs/file_formats.md
```diff
@@ -9,8 +9,9 @@ When you are ready to use Casanovo, you can input your MS/MS spectra in one of t
 - **[mzML](https://doi.org/10.1074/mcp.R110.000133)**: XML-based mass spectrometry community standard file format developed by the Proteomics Standards Initiative (PSI).
 - **[mzXML](https://doi.org/10.1038/nbt1031)**: XML-based predecessor of mzML. Although supported by Casanovo, mzML should typically be preferred instead.
 - **[MGF](https://www.matrixscience.com/help/data_file_help.html)**: A simple text-based peak file, though not as rich in detail as mzML.
+- **[TDF](https://doi.org/10.1016/j.mcpro.2021.100149)**: A proprietary, binary file format that contains a relational database **(.tdf)** for metadata and **(.tdf_bin)** for the intensity data.
```
bittremieux: **question:** The code doesn't support a `.tdf` extension, but rather `.d`. Where does `.tdf` suddenly come from?

tests/unit_tests/test_runner.py
```diff
@@ -14,6 +14,31 @@
...
+def test_loading_timstof_folders(tiny_config, tmp_path, mgf_small):
+    config = Config(tiny_config)
+    config.max_epochs = 1
+    config.n_layers = 1
+    ckpt = tmp_path / "test.ckpt"
+
+    real_path = tmp_path / "sample.d"
+    real_path.mkdir()
+
+    with ModelRunner(config=config, output_dir=tmp_path) as runner:
```
bittremieux: **question:** Why is training a model needed for this test? It will just make it run slower.

You might even just mock the `ModelRunner`, because its functionality is not needed to test the input path functionality.

tests/unit_tests/test_runner.py
```diff
@@ -14,6 +14,31 @@
...
+        runner.train([mgf_small], [mgf_small])
+        runner.trainer.save_checkpoint(ckpt)
+
+    paths = runner._get_input_paths(
+        peak_path=(real_path,), annotated=False, mode="train"
+    )
+    assert len(paths) > 0
```
bittremieux: **todo:** Check should be more specific.

tests/unit_tests/test_runner.py
```diff
@@ -14,6 +14,31 @@
...
+
+    fake_path = tmp_path / "nonexistent"
```
bittremieux: **todo:** This is different because of two reasons:
1. The path is fake.
2. It doesn't have a supported extension.

This test now can't differentiate between those two aspects.

CHANGELOG.md
```diff
@@ -34,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Training and optimizer metrics can be logged to a CSV file by setting the `log_metrics` config file option to true. The CSV file will be written to under a sub-directory of the output directory named `csv_logs`.
 - New configuration options for detailed control of the gradients during training (gradient accumulation, clipping).
 - New configuration option `min_peaks` to discard low-quality spectra with too few peaks.
+- timsTOF files can now be accepted as spectra input files
```
bittremieux: **nitpick:** End sentence with period.

casanovo/denovo/model_runner.py
```diff
@@ -502,7 +502,10 @@ def initialize_model(self, train: bool, db_search: bool = False) -> None:
         model_clss = DbSpec2Pep if db_search else Spec2Pep
         try:
             self.model = model_clss.load_from_checkpoint(
-                self.model_filename, map_location=device, **loaded_model_params
+                self.model_filename,
+                map_location=device,
+                **loaded_model_params,
+                weights_only=False,
```
bhavyak7-lab, line 518: > Unsupported global: GLOBAL numpy.core.multiarray.scalar was not an allowed global by default.

The most recent version of pytorch lightning (2.6.0, which is what is installed with the tests) defaults the weights_only parameter to True so it doesn't allow the loading of files that do not have accepted globals. I think that the checkpoint files that Casanovo saves an unallowed property. We can override this by doing weights_only=False because it will allow all checkpoint files will be loaded in, but I don't know if this would be the best solution.

bittremieux: I don't really understand what you're saying, `weights_only` is not a global variable. We also explicitly don't want to load the weights only, but the full training history.

Please work with @Lilferrit to diagnose and fix the exact issue.

tests/unit_tests/test_runner.py
```diff
@@ -14,6 +14,41 @@
...
+
+    monkeypatch.setattr(ModelRunner, "__init__", minimal_init)
+    runner = ModelRunner()
+
+    # Testing real path
+    real_path = tmp_path / "sample.d"
+    real_path.mkdir()
+
+    paths = runner._get_input_paths(
+        peak_path=(str(real_path),), annotated=False, mode="train"
```
bittremieux: **todo:** This doesn't make sense. .d files are not annotated data, and thus can't be used for training.

tests/unit_tests/test_runner.py
bittremieux: **todo:** Should test for `length == 1`.

tests/unit_tests/test_runner.py
```diff
@@ -14,6 +14,41 @@
...
+    )
+
+    assert len(paths) > 0
+    assert paths == [str(real_path)]
```
bittremieux: **todo:** I don't think that this is a very good test, because list memory locations will actually not be the same and this unit test might fail. Instead it should check for the element in the list (in combination with verifying that there was a single element in the list in the line above).

tests/unit_tests/test_runner.py
```diff
@@ -14,6 +14,41 @@
...
+
+    # Testing unsupported extension
+    fake_path = tmp_path / "test.hi"
+    fake_path.mkdir()
+    with pytest.raises(FileNotFoundError):
+        runner._get_input_paths(
+            peak_path=(fake_path,), annotated=False, mode="train"
```
bittremieux: **todo:** Same unannotated–train inconsistency.

tests/unit_tests/test_runner.py
bittremieux: **todo:** This still needs to be addressed.

bittremieux (changes requested):

bittremieux:
This looks good, thanks @bhavyak7-lab. We'll wait a bit longer to merge it into dev because there needs to be a new release of DepthCharge first, which we'll need to specify in the dependencies then.

PR #557, opened 2025-11-15
Minumum Peptide Length Fix

Lilferrit:
As it turns out this bug is actually caused by the current (which I'm not sure if this is what we actually want) behavior of marking beams that exceed the precursor ppm tolerance as finished but not discarded. If a beam exceeds the precursor tolerance but has a length less than the minimum peptide length, it will be marked as finished and then subsequently cached.

However, I am not convinced the current behavior of marking beams as finished but not discarded if they exceed the precursor tolerance is actually what we want, because this will lead to predictions that exceed the precursor tolerance, sometimes by quite a bit, to be written to the output mztab. I wrote the fix assuming this is the behavior that we want, but I could fairly easily update it to discard any beams that exceed the precursor ppm tolerance. I will also update the unit test(s) once we decide on the behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,19 +19,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - A mismatching parameter warning will now only be triggered for the tokenizer if the config and checkpoint tokenizers do not have equivalent vocabularies.
 - Removed erroneous tokenizer vocabulary warning.
 - Fixed an issue which led the reported peptide precision to be 0 during evaluation mode.
-- Peptide predictions that are below the minimum peptide length but are within the precursor mass tolerance will no longer be written to the output.
+- Peptide predictions failing the minimum peptide length are not reported, irrespective of whether they match or exceed the precursor mass.
 
 ## [5.1.2] - 2025-12-11
 
 ### Changed
 
 - Hotfix to ensure compatibility with PyTorch v2.6 when loading weights files.
-
-## [5.1.1] - 2025-11-27
-
-### Changed
-
-- Hotfix to avoid excessively slow peptide parsing during mzTab export.
 
 ## [5.1.1] - 2025-11-27
 


wsnoble:
Can you run Casanovo with and without this fix on the MassIVE-KB test set and then pull out the predictions that differ?  I'd be interested to see old and new scores and predictions, as well as an indication of which ones are correct.

Lilferrit:
> Can you run Casanovo with and without this fix on the MassIVE-KB test set and then pull out the predictions that differ? I'd be interested to see old and new scores and predictions, as well as an indication of which ones are correct.

I ran it on the full test set with and without the fix and there were no predictions that differed. I suppose it makes sense that this edge case is pretty rare, as you would need a really low precursor mass in order to get a prediction that simultaneously exceeds the precursor mass filter and is below the (default) minimum peptide length. It would however be possible to trigger the edge case for testing purposes by setting the minimum peptide length to something like 20, which could be a nice sanity check.

bittremieux:
Yes, we should implement the fix and add such a test.

Lilferrit:
The diff for this is quite large but most of that is just coming from the refactored unit tests.

CHANGELOG.md
```diff
@@ -32,6 +33,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Hotfix to avoid excessively slow peptide parsing during mzTab export.
 
+## [5.1.1] - 2025-11-27
```
bittremieux: **todo:** This is already in the changelog on the lines above, remove.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - A mismatching parameter warning will now only be triggered for the tokenizer if the config and checkpoint tokenizers do not have equivalent vocabularies.
 - Removed erroneous tokenizer vocabulary warning.
 - Fixed an issue which led the reported peptide precision to be 0 during evaluation mode.
+- Peptide predictions that are below the minimum peptide length but are within the precursor mass tolerance will no longer be written to the output.
```
bittremieux: **nitpick:** Simplify as "Peptide predictions failing the minimum peptide length are not reported, irrespective of whether they match or exceed the precursor mass."

tests/unit_tests/test_unit.py
bittremieux, line 1: **question:** It's a bit hard to ascertain because the diff is not very useful, but I assume that all test cases are still represented?

Lilferrit: Yep exactly, 90% of the test case code is reused except for some additional test cases and using `pytest.mark.parameterize` to eliminate structural redundancy for some of the test cases. All of the original test cases are still present.

PR #563, opened 2025-12-08
Specify weights only keyword

Lilferrit:
In Pytorch 2.6 they changed the default value of the `weights_only` for `Torch.load` from `False` to `True`, which breaks our model loading implementation. I'm basing this against `main` from now under the assumption that this will be a hotfix, but I can rebase it to `dev` if we want to push this to a later release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


bittremieux (commented):
This looks fine as a hotfix, but we need to get the unit tests working again first.

Lilferrit:
Sounds good - I found what was causing that unit test to fail.

bittremieux (dismissed):

bittremieux (approved):

PR #574, opened 2026-02-12
Model weights loading

### LINKED ISSUE #570: Resuming training from checkpoint not possible
## Describe the Issue
When resuming training from a given checkpoint using `casanovo -train --model pathtomodel.ckpt traindata.mgf`, not the whole checkpoint (e.g. LR scheduler etc.) is loaded but only the weights. This means the learning rate is reset, resulting in weird behaviour in the subsequent training process. 
Training a model for x hours, stopping, then resuming training from the latest checkpoint is a common workflow e.g. in HPC settings, where jobs are limited to for example 24 hours.

## Steps To Reproduce
Train a model from scratch, stop training, resume training with the command above.

## Expected Behavior
Both should be possible: 
(1) Loading the complete checkpoint and (2) only the weights.
(2) is still helpful for e.g. finetuning Casanovo on a smaller dataset, where using the original LR scheduler might hurt.
Technically, it's a small fix, maybe add a flag in the config to choose between (1) and (2). For (1), add `ckpt_path=self.model_filename` to the trainer.fit call.

bhavyak7-lab:
Modified config file to have a flag on whether or not to load all parts of the checkpoint or only the weights, and altered the model_runner code to load the checkpoint differently based on this flag. Addresses #570 

## Summary by CodeRabbit

* **New Features**
  * Added --load_all_states CLI flag to resume training and restore full model state.

* **User-facing Validation**
  * Warns if resuming without a model; errors if a model is a URL or the checkpoint file is missing.

* **Tests**
  * Integration test exercises resume-training flow; unit test verifies validation messages and errors.

* **Documentation**
  * Changelog entry describing resume behavior.

* **Chores**
  * Bumped minimum Lightning requirement to 2.6.


casanovo/config.yaml
```diff
@@ -148,6 +148,10 @@ num_sanity_val_steps: 0
 # Calculate peptide and amino acid precision during training.
 # This is expensive, so we recommend against it.
 calculate_precision: false
+# When restarting training from a checkpoint:
+# - true: Load only model weights (start fresh optimizer/scheduler/epoch)
+# - false: Resume full training state (optimizer, scheduler, epoch from checkpoint)
+load_weights_only: true
```
Lilferrit: IMO it would be more intuitive to make this a command line argument for the train sub-command rather than a config file option. I.e. setting `--load_weights_only` would not load the optimizer state when starting a new training run.

I also think the code rabbit suggestion of popping a warning if this flag is set but no checkpoint file is provided would be nice too.

casanovo/denovo/model_runner.py
```diff
@@ -198,6 +204,8 @@ def train(
             self.model,
             self.loaders.train_dataloader(),
             self.loaders.val_dataloader(),
+            ckpt_path=ckpt_path,
+            weights_only=False,
```
Lilferrit: The user argument for `weights_only` (e.g. the CLI flag argument I talked about in another comment) should probably be propagated here rather than just setting this to always be false. There may be cases where a user may want to start training from a pre-trained set of weights, but without starting from the same optimizer state.

Lilferrit (commented):
Looks good; just a couple of minor things I think we should look at.

bittremieux:
@bhavyak7-lab Please fix linting issues, address and resolve Coderabbit comments, and tag me when the PR is ready for final review.

TheBobBob:
@bittremieux The PR is ready for review.

casanovo/denovo/model_runner.py
```diff
@@ -195,11 +199,20 @@ def train(
...
+                self.loaders.val_dataloader(),
+            )
+        else:
+            self.trainer.fit(
+                self.model,
+                self.loaders.train_dataloader(),
+                self.loaders.val_dataloader(),
+                ckpt_path=ckpt_path,
+                weights_only=False,
+            )
```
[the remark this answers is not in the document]
bittremieux: **todo:** Set the minimum Lightning version to 2.6 in the `pyproject.toml` file.

casanovo/casanovo.py
```diff
@@ -347,7 +361,13 @@ def train(
         for peak_file in validation_peak_path:
             logger.info("  %s", peak_file)
 
-        runner.train(train_peak_path, validation_peak_path)
+        if load_all_states:
```
bittremieux: **nitpick:** Can be simplified a little bit as:
```
runner.train(
    train_peak_path,
    validation_peak_path,
    model if load_all_states else None
)
```

casanovo/casanovo.py
```diff
@@ -381,6 +401,27 @@ def configure(
     logger.info(f"Wrote {config_path}")
 
 
+def _warn_if_invalid(model: Optional[str], load_all_states: bool):
```
bittremieux: **todo:** Add a docstring for this function.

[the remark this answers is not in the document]
bittremieux: **todo:** Do this.

casanovo/casanovo.py
```diff
@@ -381,6 +401,27 @@ def configure(
...
+    if load_all_states:
+        if model is None:
+            warnings.warn(
```
bittremieux: **todo:** These should use `logger.warning` instead of the warnings module. The module is used for developer warnings, whereas logging is used for user warnings. This is the latter.

casanovo/casanovo.py
```diff
@@ -381,6 +401,27 @@ def configure(
...
+                "When --load_all_states is specified, --model must also be provided."
+                "Training will resume without model information.",
```
bittremieux: **nitpick:** "Resume" is not really the correct term here, right? Training will start from scratch without a provided model.

casanovo/casanovo.py
bittremieux: **todo:** Use a more informative function name. Which validity is checked? I recommend `_is_valid_model`.

casanovo/casanovo.py
```diff
@@ -381,6 +401,27 @@ def configure(
...
+                stacklevel=2,
+            )
+        elif _is_valid_url(model):
+            warnings.warn(
```
bittremieux: **question:** This should then raise an error that halts execution, right?

casanovo/casanovo.py
```diff
@@ -381,6 +401,27 @@ def configure(
...
+                "Full model state cannot be loaded from a URL. "
+                "Please provide a local file path when --load_all_states is True.",
+                stacklevel=2,
+            )
+        elif not Path(model).is_file():
+            warnings.warn(
+                "When --load_all_states is True, the model path must point to an existing file.",
+                stacklevel=2,
+            )
```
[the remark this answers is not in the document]
bittremieux: **todo:** Fix this.

tests/unit_tests/test_unit.py
```diff
@@ -264,6 +264,25 @@ def test_version():
     assert casanovo.__version__ is not None
 
 
+def test_warn_if_invalid():
```
bittremieux: **suggestion:** Maybe slightly cleaner using `pytest.mark.parametrize`.

tests/test_integration.py
```diff
@@ -49,6 +49,42 @@ def test_train_and_run(
     assert model_file.exists()
     assert best_model.exists()
 
+    with tiny_config.open("r") as f:
+        config_data = yaml.safe_load(f)
+    original_max_epochs = config_data.get("max_epochs")
+
+    try:
+        config_data["max_epochs"] = 40
```
bittremieux: **question:** Why does this need to be increased?

tests/test_integration.py
```diff
@@ -49,6 +49,42 @@ def test_train_and_run(
...
+            "--output_dir",
+            str(tmp_path),
+            "--output_root",
+            "train_resuming",
+            "--model",
+            str(model_file),
+            "--load_all_states",
+        ]
+
+        result = run(train_args)
```
bittremieux, line 80: **question:** I only see a single `train` here, where does the resuming happen?

bittremieux (changes requested):
Please address the comments and/or answer the questions. Also update the changelog.

tests/test_integration.py
bhavyak7-lab: Lightning does not restart the training with the fully loaded model state if it had reached max_epochs in its initial training, so I had to increase the max_epochs so I could test whether or not the training began correctly.

tests/test_integration.py
bhavyak7-lab: The initial training was completed above from lines 34-54, and I am using that same model file to resume the training.

bittremieux (approved):

PR #579, opened 2026-03-02
Fix minimum length off by one bug

Lilferrit:
This bug was caused by this code block:

```python
if self.tokenizer.reverse:
    has_stop_at_start = tokens[:, 0] == self.stop_token
    peptide_lens[has_stop_at_start] -= 1
else:
    has_stop_at_end = ends_stop_token
    peptide_lens[has_stop_at_end] -= 1
```

Regardless of the decoding direction stop tokens are always predicted at the end of their corresponding vector in the tokens tensor. However, in reverse mode this block looks for stop tokens at the beginning of the token vectors, hence it doesn't properly adjust the length for the stop token.

## Summary by CodeRabbit

* **Bug Fix**
  * Fixed an issue where some predicted peptides were reported one residue shorter than the configured minimum; these predictions are now correctly filtered.

* **Refactor**
  * Simplified internal peptide-length handling during beam finishing to remove special-case behavior and make length decisions more consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fixed an issue which led the reported peptide precision to be 0 during evaluation mode.
 - Peptide predictions failing the minimum peptide length are not reported, irrespective of whether they match or exceed the precursor mass.
 - Setting `--output_root` to a directory will no longer cause an error.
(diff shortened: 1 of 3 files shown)


bittremieux (approved):
Thanks for catching this bug! Please still add a note to the changelog.

bittremieux (commented):
@Lilferrit Can you resolve the merge conflicts and add a note to this one to merge it? There are some minor conflicts in related beam search changes.

PR #582, opened 2026-03-07
Feature/track mgf scan numbers

### LINKED ISSUE #142: Track scan numbers in MGF files
It is unfortunate that scan numbers in MGF files are not retained by Casanovo.  We should make it possible to keep this information, in addition to the scan index. In the future, DepthCharge is going to be modified to be more flexible with the information it parses from the mass spec data files. When that happens, we can use that functionality to track scan numbers.  However, prior to these changes, we could accomplish this by adding to Casanovo a post-processing step wherein the MGF file is read again. It’s not hard to do---something like the following would work:
```
with open("test.mgf") as mgf_file:
    index_counter = 0 
    index2scan = {}
    for line in mgf_file:
        if line.startswith("SCAN ID"):
            index2scan[index_counter] = line.split("=", 1)[1]
            index_counter += 1
```

`index2scan` would be a dictionary mapping the scan index to the scan id value, again assuming all scans had the `SCAN ID` field.

In practice, we should probably support `SCAN ID`, `SCAN` and `SCANS`, since there is no definitive standard for MGF.

GauravSRC:
## Summary

Closes #142.

MGF files can carry instrument-assigned scan numbers in the `SCANS`,
`SCAN`, or `SCAN ID` header field. This PR threads those values through
the pipeline so they appear in the mzTab `spectra_ref` column alongside
the existing file index, e.g.:

    ms_run[1]:index=0 scan=17

## Approach

The implementation reads only the header lines of each MGF spectrum
entry (never the peak data) to build a lightweight index mapping
spectrum position to scan number. This index is pre-computed once in
`MztabWriter` before prediction begins and looked up at write time --
no second full file read is required during the hot path.

This approach was chosen over the DepthCharge `CustomField` mechanism
because that mechanism stores `None` for mzML spectra and for MGF
files that omit the `SCANS` field, which causes an uncaught
`RuntimeError` in DepthCharge 0.4.9's `_tensorize` function when
those null values reach `torch.tensor()`.

## Files changed

| File | Change |
|---|---|
| `casanovo/data/ms_io.py` | Add `_build_mgf_scan_index()` helper and `MztabWriter.set_mgf_scan_index()`; embed scan number in `spectra_ref` inside `save()` |
| `casanovo/denovo/model_runner.py` | Call `set_mgf_scan_index()` in both `predict()` and `db_search()` |
| `tests/conftest.py` | `_create_mgf` / `_create_mgf_entry` accept optional `scan_nums`; two new fixtures |
| `tests/unit_tests/test_unit.py` | 5 new unit tests |

## Backward compatibility

- MGF files without a `SCANS` field: output unchanged (`index=N` only).
- mzML files: completely unaffected.
- All existing integration tests pass.

## Summary by CodeRabbit

* **New Features**
  * Export instrument-assigned scan numbers from MGF headers (SCANS/SCAN/SCAN ID) into mzTab as an optional global column (ms_run[<i>]:scan=<n>); shown when any MGF input includes scan numbers, with nulls for other spectra in mixed runs. MGF spectrum references are normalized to preserve consistent linking.

* **Tests**
  * Added fixtures and unit/integration tests covering MGF scan-header extraction and mzTab output across MGF/mzML combinations.

* **Documentation**
  * Updated file-format docs and changelog (repositioned a release entry).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,17 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
-- Track instrument-assigned scan numbers from MGF `SCANS`, `SCAN`, and `SCAN ID` header fields in a new `opt_global_cv_MS:1003057_scan_number` mzTab column.
-
-## [5.1.2] - 2025-12-11
-
-### Changed
-
-- Hotfix to ensure compatibility with PyTorch v2.6 when loading weights files.
-
-### Added
-
 - A TSV file with all candidate peptides can be exported during database searching with the `--export` flag.
+- Track instrument-assigned scan numbers from MGF `SCANS`, `SCAN`, and `SCAN ID` header fields in a new `opt_global_cv_MS:1003057_scan_number` mzTab column.
 
 ### Changed
 
@@ -34,6 +25,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fixed an issue which led the reported peptide precision to be 0 during evaluation mode.
 - Peptide predictions failing the minimum peptide length are not reported, irrespective of whether they match or exceed the precursor mass.
 - Setting `--output_root` to a directory will no longer cause an error.
+
+## [5.1.2] - 2025-12-11
+
+### Changed
+
+- Hotfix to ensure compatibility with PyTorch v2.6 when loading weights files.
 
 ## [5.1.1] - 2025-11-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
-- Track instrument-assigned scan numbers from MGF file `SCANS`, `SCAN`, or
-  `SCAN ID` header fields and include them in the mzTab `spectra_ref` column
-  alongside the spectrum file index (e.g. `ms_run[1]:index=0 scan=17`).
-
+- Track instrument-assigned scan numbers from MGF `SCANS`, `SCAN`, and `SCAN ID` header fields in a new `opt_global_cv_MS:1003057_scan_number` mzTab column.
 ## [5.1.2] - 2025-12-11
 
 ### Changed
@@ -39,9 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
-- Track instrument-assigned scan numbers from MGF file `SCANS`, `SCAN`, or
-  `SCAN ID` header fields and include them in the mzTab `spectra_ref` column
-  alongside the spectrum file index (e.g. `ms_run[1]:index=0 scan=17`).
+- Track instrument-assigned scan numbers from MGF `SCANS`, `SCAN`, and `SCAN ID` header fields in a new `opt_global_cv_MS:1003057_scan_number` mzTab column.
 
 ## [5.1.2] - 2025-12-11
 
@@ -362,7 +357,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 [Unreleased]: https://github.com/Noble-Lab/casanovo/compare/v5.1.2...HEAD
 [5.1.2]: https://github.com/Noble-Lab/casanovo/compare/v5.1.1...v5.1.2
-[Unreleased]: https://github.com/Noble-Lab/casanovo/compare/v5.1.1...HEAD
 [5.1.1]: https://github.com/Noble-Lab/casanovo/compare/v5.1.0...v5.1.1
 [5.1.0]: https://github.com/Noble-Lab/casanovo/compare/v5.0.0...v5.1.0
 [5.0.0]: https://github.com/Noble-Lab/casanovo/compare/v4.3.0...v5.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 
 - Track instrument-assigned scan numbers from MGF `SCANS`, `SCAN`, and `SCAN ID` header fields in a new `opt_global_cv_MS:1003057_scan_number` mzTab column.
+
 ## [5.1.2] - 2025-12-11
 
 ### Changed
@@ -33,16 +34,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - Fixed an issue which led the reported peptide precision to be 0 during evaluation mode.
 - Peptide predictions failing the minimum peptide length are not reported, irrespective of whether they match or exceed the precursor mass.
 - Setting `--output_root` to a directory will no longer cause an error.
-
-### Added
-
-- Track instrument-assigned scan numbers from MGF `SCANS`, `SCAN`, and `SCAN ID` header fields in a new `opt_global_cv_MS:1003057_scan_number` mzTab column.
-
-## [5.1.2] - 2025-12-11
-
-### Changed
-
-- Hotfix to ensure compatibility with PyTorch v2.6 when loading weights files.
 
 ## [5.1.1] - 2025-11-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Added
+
+- Track instrument-assigned scan numbers from MGF file `SCANS`, `SCAN`, or
+  `SCAN ID` header fields and include them in the mzTab `spectra_ref` column
+  alongside the spectrum file index (e.g. `ms_run[1]:index=0 scan=17`).
+
 ## [5.1.2] - 2025-12-11
 
 ### Changed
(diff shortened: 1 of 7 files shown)


bittremieux:
@GauravSRC thank you for the contribution. Can you please do the following?
1. Rebase this PR against `dev` (instead of `main`).
2. Address all CodeRabbit comments.
3. After having done that, tag me for a review.

GauravSRC:
> @GauravSRC thank you for the contribution. Can you please do the following?
> 
> 1. Rebase this PR against `dev` (instead of `main`).
> 2. Address all CodeRabbit comments.
> 3. After having done that, tag me for a review.

Hello @bittremieux, thank you for reviewing this PR and for the detailed feedback -.
I have addressed all the CodeRabbit comments (removed the unused function, fixed PEP 8 blank lines, added the changelog entry, corrected the duplicate link label, and removed the redundant local imports) and rebased the branch onto dev. All existing integration tests and the five new unit tests pass cleanly.
Please do let me know if any further changes are needed.

casanovo/data/ms_io.py
```diff
@@ -8,12 +8,65 @@
...
+
 import natsort
 
+logger = logging.getLogger("casanovo")
+
 from .. import __version__
 from ..config import Config
 from .psm import PepSpecMatch
 
+def _build_mgf_scan_index(mgf_path: str) -> dict:
```
bittremieux: **nitpick:** Include type hints for the dictionary keys and values.

casanovo/data/ms_io.py
```diff
@@ -8,12 +8,65 @@
...
+        with open(mgf_path, "r", errors="replace") as fh:
+            for line in fh:
+                stripped = line.strip()
+                upper = stripped.upper()
+                if upper == "BEGIN IONS":
+                    in_ions = True
+                    current_scan = None
+                elif upper == "END IONS":
+                    if current_scan is not None:
+                        index2scan[str(index)] = current_scan
```
bittremieux: **suggestion:** Rather than storing the mapping in a full dict here and then iterating over it in line 220, you could `yield` the (index, scan) tuple here. It might be a little bit more efficient.

casanovo/data/ms_io.py
```diff
@@ -185,6 +261,21 @@ def save(self) -> None:
                 1,
             ):
                 filename, idx = psm.spectrum_id
+                if Path(filename).suffix.lower() == ".mgf":
+                    # Normalize idx: handle both "0" and "index=0" shapes.
```
bittremieux: **todo:** A format `"0"` should never occur, it should also have the index prefix. This can then directly be incorporated in the `_mgf_scan_index` so that you can avoid the string indexing stuff.

casanovo/data/ms_io.py
```diff
@@ -185,6 +261,21 @@ def save(self) -> None:
...
+                        idx[len("index="):]
+                        if idx.startswith("index=")
+                        else idx
+                    )
+                    if numeric_idx.isnumeric():
+                        scan_num = self._mgf_scan_index.get(
+                            (filename, numeric_idx)
+                        )
+                        idx = f"index={numeric_idx}"
+                        if scan_num:
```
bittremieux: **todo:** This is an incorrect way to specify the scan number and breaks mzTab compatibility. Please check the mzTab specification here: https://github.com/HUPO-PSI/mzTab/blob/master/specification_document-releases/1_0-Proteomics-Release/mzTab_format_specification.pdf

TLDR, see page 7:
- MGF file indexes should be specified with `index=xsd:nonNegativeInteger`. This is the only official supported spectrum identifier for MGF files.
- We can add the _unofficial_ scan number, but this can't be in the same column. Instead, we can use a new, optional column with header `opt_global_cv_MS:1003057_scan_number`, which should take values of the type `ms_run[1]:scan=2` (with the correct run index).

casanovo/denovo/dataloaders.py
```diff
@@ -193,6 +192,11 @@ def _make_dataset(
         torch.utils.data.Dataset
             A PyTorch Dataset for the given peak files.
         """
+        # Include the MGF scan number CustomField only for test/inference
```
bittremieux: **todo:** I don't know what the relevance of this comment is, because there's no code change and the current implementation of the custom field doesn't cover a SCANS key.

casanovo/denovo/model_runner.py
```diff
@@ -164,6 +164,7 @@ def db_search(
         )
         test_paths = self._get_input_paths(peak_path, False, "test")
         self.writer.set_ms_run(test_paths)
```
bittremieux, line 166: **suggestion:** Create the index-to-scan mapping implicitly in `ms_io` rather than having to do it explicitly in multiple locations in the `model_runner`.

tests/unit_tests/test_unit.py
```diff
@@ -2522,3 +2522,154 @@ def dummy(*args, **kwargs):
         assert res.exit_code == 0
         assert "Error" not in res.stderr
         assert "ok" in res.stdout
+# issue 142- mgf scan number tracking
```
bittremieux: **nitpick:** Remove comment.

tests/unit_tests/test_unit.py
```diff
@@ -2522,3 +2522,154 @@ def dummy(*args, **kwargs):
...
+    result = _build_mgf_scan_index(str(mgf_file))
+    assert result == {"0": "42"}
+
+
+def test_scan_num_in_mztab_spectra_ref(tmp_path):
+    """
+    After set_mgf_scan_index is called, MztabWriter must embed the scan
+    number in spectra_ref: 'ms_run[1]:index=0 scan=17'.
+    """
+
```
bittremieux: **nitpick:** Excessive empty lines.

tests/unit_tests/test_unit.py
```diff
@@ -2522,3 +2522,154 @@ def dummy(*args, **kwargs):
...
+    assert "index=0" in spectra_ref, f"Missing index=0 in: {spectra_ref}"
+    assert "scan=17" in spectra_ref, f"Missing scan=17 in: {spectra_ref}"
+
+
+def test_no_scan_num_keeps_index_only_spectra_ref(tmp_path):
+    """
+    An MGF without SCANS must produce 'ms_run[1]:index=0' only,
+    preserving backward compatibility.
+    """
+
```
bittremieux: **nitpick:** Excessive empty lines.

tests/unit_tests/test_unit.py
```diff
@@ -2522,3 +2522,154 @@ def dummy(*args, **kwargs):
...
+
+def test_mzml_spectra_ref_unaffected_by_scan_num_feature(
+    mzml_small, tmp_path
+):
+    """
+    mzML spectra use their native scan ID and must be unchanged even
+    when set_mgf_scan_index is called (mzML files are silently ignored).
+    """
+
```
bittremieux: **nitpick:** Excessive empty lines.

tests/unit_tests/test_unit.py
bittremieux, line 1: **praise:** Nice job adding comprehensive unit tests.

CHANGELOG.md
```diff
@@ -8,6 +8,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+- Track instrument-assigned scan numbers from MGF file `SCANS`, `SCAN`, or
```
bittremieux: **todo:** Fix duplicate changelog entries. Put it on a single line.

CHANGELOG.md
```diff
@@ -344,6 +362,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 [Unreleased]: https://github.com/Noble-Lab/casanovo/compare/v5.1.2...HEAD
 [5.1.2]: https://github.com/Noble-Lab/casanovo/compare/v5.1.1...v5.1.2
+[Unreleased]: https://github.com/Noble-Lab/casanovo/compare/v5.1.1...HEAD
```
bittremieux: **todo:** Fix.

bittremieux (changes requested):
The most important comment is that the new scan numbers should be compatible with the official mzTab standard. Please see details below.

It would be a bit more satisfying if the scan number field would be directly read by the `MgfParser` in depthcharge, but this workaround is probably a lot simpler.

Also please make sure that the linting issues are fully resolved.

GauravSRC:
Hii @bittremieux!
I have addressed all your review comments - type hints and generator refactor for _build_mgf_scan_index, scan number moved to the new opt_global_cv_MS:1003057_scan_number column per the mzTab spec, index building folded into set_ms_run, stale comments removed, blank lines and CHANGELOG fixed, and Black/ruff pass cleanly. All 9 tests pass. Thank you for the thorough and helpful review! - I learned a lot from it. Please let me know if anything else needs attention.

bittremieux (changes requested):
Thanks for the changes, the code looks good.

I'd like to request one more update: in the [documentation on the mzTab output](https://github.com/Noble-Lab/casanovo/blob/main/docs/file_formats.md#output-understanding-the-mztab-format) there's info on the `spectra_ref` column for various types of input files. Please add some information about the newly added column with scan information from the supported header keys for MGF files.

GauravSRC:
> Thanks for the changes, the code looks good.
> 
> I'd like to request one more update: in the [documentation on the mzTab output](https://github.com/Noble-Lab/casanovo/blob/main/docs/file_formats.md#output-understanding-the-mztab-format) there's info on the `spectra_ref` column for various types of input files. Please add some information about the newly added column with scan information from the supported header keys for MGF files.

Hi @bittremieux ! I have addressed your requests. The opt_global_cv_MS:1003057_scan_number column is now conditional - it only appears in the mzTab output when at least one MGF input file contains scan number headers; mzML, mzXML, and scan-less MGF runs produce output identical to before. I have also added documentation to docs/file_formats.md describing the new column, which header keys populate it, its format, and when it is present versus absent.

casanovo/data/ms_io.py
```diff
@@ -2,18 +2,75 @@
...
+    tuple[str, str]
+        A ``(spectrum_ref_id, scan_number)`` pair where
+        *spectrum_ref_id* uses the ``index=N`` zero-based MGF index.
+    """
+    index, current_scan, in_ions = 0, None, False
+    try:
+        with open(mgf_path, errors="replace") as fh:
+            for line in fh:
+                upper = line.strip().upper()
+                if upper == "BEGIN IONS":
```
bittremieux: I'd opt to avoid using magic strings like this inside your code, but predefine them as variables at the top of your code

bittremieux (approved):

bittremieux (approved):

PR #586, opened 2026-03-12
Added control frequency of reporting training loss

### LINKED ISSUE #585: Allow user to control frequency of reporting training loss
We should add to the config file an option called `train_check_interval` that is analogous to the existing `val_check_interval`. Then do the loss reporting by step instead of by epoch in model.py and give this config option to the Trainer in model_runner.py.

mhmdkardosha:
## Summary
This PR adds `train_check_interval` as a config option (analogous to `val_check_interval`) and switches training loss reporting from epoch-based to step-based logging.

## What Changed

1. Added `train_check_interval` to configuration
- Added `train_check_interval` to `casanovo/config.yaml`.
- Registered `train_check_interval` in config type parsing in `casanovo/config.py`.

2. Updated model logging to be step-based
- In `casanovo/denovo/model.py`:
  - Replaced `n_log` with `train_check_interval` in `Spec2Pep`.
  - Changed loss logging behavior in `training_step`:
    - training loss logs on step (`on_step=True` for train)
    - validation loss logs on epoch (`on_epoch=True` for valid)
  - Replaced `on_train_epoch_end` with `on_train_batch_end` for training-loss history logging.
  - Gated train-history logging and console output by `train_check_interval`.

3. Wired interval into Trainer and model construction
- In `casanovo/denovo/model_runner.py`:
  - Passed `log_every_n_steps=self.config.train_check_interval` to Lightning `Trainer`.
  - Passed `train_check_interval=self.config.train_check_interval` into both
    - `model_params`
    - `loaded_model_params`

4. Updated tests/config fixtures
- In `tests/conftest.py`, added `train_check_interval` to test config defaults.
- In `tests/unit_tests/test_runner.py`, updated metrics logging test to set `config.train_check_interval = 1`.

## Why
It solves #585, as previously, training loss reporting frequency was controlled by epoch-oriented behavior (`n_log`) and did not align with the requested step-based control. This change makes training loss logging explicitly step-based and configurable via `train_check_interval`, matching `val_check_interval` semantics.

## Notes
It was made with the help of AI.

## Summary by CodeRabbit

* **New Features**
  * Added `train_check_interval` configuration parameter to control training progress logging frequency (default: 50,000 steps).

* **Chores**
  * Deprecated `n_log` parameter; use `train_check_interval` instead in configuration files.
  * Updated documentation examples and test configurations to reflect new parameter name.


casanovo/denovo/model.py
```diff
@@ -110,7 +109,7 @@ def __init__(
         min_peptide_len: int = 6,
         n_beams: int = 1,
         top_match: int = 1,
-        n_log: int = 10,
+        train_check_interval: int = 50_000,
         train_label_smoothing: float = 0.01,
```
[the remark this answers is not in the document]
bittremieux: **todo:** This is a relevant comment, please address it.

casanovo/denovo/model.py
```diff
@@ -1079,8 +1078,8 @@ def training_step(
         self.log(
             f"{mode}_CELoss",
             loss.detach(),
-            on_step=False,
-            on_epoch=True,
+            on_step=(mode == "train"),
+            on_epoch=(mode != "train"),
```
bittremieux, line 836: **question:** Why the difference between training vs not? The validation logging interval is also expressed in steps, rather than epochs.

casanovo/denovo/model.py
```diff
@@ -1185,10 +1184,12 @@ def predict_step(
 
         return predictions
 
-    def on_train_epoch_end(self) -> None:
+    def on_train_batch_end(self, outputs, batch, batch_idx) -> None:
```
bittremieux: **todo:** Change function signature to reflect that parameters are not used.

casanovo/denovo/model.py
```diff
@@ -947,10 +947,12 @@ def predict_step(
...
         """
-        Log the training loss at the end of each epoch.
+        Log the training loss at the end of each training step.
         """
+        if self.trainer.global_step % self.train_check_interval != 0:
```
bittremieux: **question:** Do we need to explicitly check for this, or does Lightning handle it for us?

tests/unit_tests/test_runner.py
```diff
@@ -375,7 +375,7 @@ def test_evaluate_failure_cases(
 def test_metrics_logging(tmp_path, mgf_small, tiny_config):
     config = Config(tiny_config)
     config.log_metrics = True
-    config.log_every_n_steps = 1
+    config.train_check_interval = 1
```
bittremieux, line 406: **suggestion:** Logging every single step is probably a bit excessive, change this to a relevant value for the context of the unit tests.

tests/conftest.py
```diff
@@ -287,6 +287,7 @@ def _get_config_file(file_path, file_name, additional_cfg=None):
         "log_metrics": False,
         "log_every_n_steps": 50,
         "lance_dir": None,
+        "train_check_interval": 1,
```
bittremieux: **todo:** Same here as above, use a relevant value.

bittremieux (changes requested):
Please see the comments.

Also add an integration test to verify that the training and/or validation loss are correctly reported after each specified number of steps.

casanovo/denovo/model.py
mhmdkardosha: @bittremieux When the validation loop runs, it makes more sense to monitor the average across the entire validation subset, rather than noisy step-by-step metrics for each individual validation batch. Also as it was before, the validation logging was expressed in epochs in the old code.

casanovo/denovo/model.py
mhmdkardosha: @bittremieux Yes, we do. Without the check, we will be recording the loss at each batch end, which we don't want. But if we want a customized history or logging (which is the case), then we need to write a condition to check it.

tests/unit_tests/test_runner.py
mhmdkardosha: @bittremieux For the context of the unit tests, the test is already small, so this value is reasonable.

tests/conftest.py
mhmdkardosha: @bittremieux Yes, this needs to be edited to a realistic value. I edited it to be 50.

mhmdkardosha:
Done @bittremieux. I addressed all your comments, and I am waiting for your review. And sorry for the delay, I was having the Eid vacation.

casanovo/denovo/model.py
```diff
@@ -108,6 +110,12 @@ def __init__(
         **kwargs: Dict,
     ):
         super().__init__()
+        if (
+            not isinstance(train_check_interval, Integral)
```
bittremieux: **todo:** Config value checks should uniformly happen in `config.py`, rather than at potentially various locations throughout the code.

casanovo/denovo/model.py
```diff
@@ -142,11 +150,13 @@ def __init__(
         # unrecognized arguments, including deprecated ones. Remove the
         # deprecated ones.
         for k in config._config_deprecated:
-            kwargs.pop(k, None)
-            warnings.warn(
-                f"Deprecated hyperparameter '{k}' removed from the model.",
-                DeprecationWarning,
-            )
+            if k in kwargs:
```
bittremieux, line 144: **question:** Why is this change needed? It seems functionally equivalent to the previous code.
Additionally, why should `n_log` be handled separately rather than be added to `config._config_deprecated`?

bittremieux (changes requested):

bittremieux:
Can you show an example of a training run that shows the updated training/validation logging frequency?

mhmdkardosha:
Thanks for the feedback @bittremieux , this makes sense.

I updated the change to keep config checks centralized and reduce special-casing:
- `train_check_interval` validation now lives in config parsing (single source of truth)
- The duplicate check in the model constructor is removed
- n_log is removed entirely (schema/default/docs/tests) and just kept in the `_config_deprecated`, so it is no longer handled as a separate model kwarg path
- kwargs cleanup in the model now only handles deprecated keys consistently

I also ran focused unit tests for config parsing and model init paths after these updates.

And I ran a train run to test the logs. I used this command:
```bash
casanovo train --config ../casanovo.yaml --output_dir . --output_root casa_experiement ../sample.mgf
```

And the files I used:
[sample.mgf](https://github.com/user-attachments/files/26305296/sample.mgf.txt)
[casanovo.yaml](https://github.com/user-attachments/files/26305297/casanovo.yaml)

And this was the result, showing that the logs are being shown as expected:
[casa_experiement.log](https://github.com/user-attachments/files/26305345/casa_experiement.log)

The config `casanovo.yaml` contained the `n_log` config option, so I ran this command to see the output, and it worked as expected:
```bash
~/casanovo python -W always::DeprecationWarning -c "from casanovo.config import Config; Config('../casanovo.yaml')"

>> OUTPUT
/teamspace/studios/this_studio/casanovo/casanovo/config.py:129: DeprecationWarning: Deprecated config option 'n_log' is no longer in use
  warnings.warn(warning_msg, DeprecationWarning)
  ```

mhmdkardosha:
Hi @bittremieux. I hope you didn't forget this :)

Lilferrit:
I'm not convinced the extra code complexity is worth the additional functionality here. Imo the best approach would be to make this mostly a config change, and allow the user to log the step-wise training loss if they wish to.

mhmdkardosha:
Hi @Lilferrit, thanks for the review. @bittremieux and I had originally discussed going with this approach, but I'm happy to change direction if a config change is preferred.

@bittremieux, I'd love to get your thoughts on this before I make any updates.

PR #588, opened 2026-03-16
fix: remove stale multiprocessing workaround in n_workers

### LINKED ISSUE #587: get_num_workers() always returns 0, disabling dataloader multiprocessing
The `return 0` on line 45 of `casanovo/utils.py` was added as a workaround for pytorch/pytorch#70344, but that issue was closed in November 2022. As a result, dataloader multiprocessing is permanently disabled for all users including Linux/GPU. The existing logic below the early return already handles platform differences correctly. Happy to remove the workaround if this looks good.

PewterZz:
Fixes #587

The early `return 0` in `n_workers()` was a workaround for pytorch/pytorch#70344, which was resolved in November 2022. Removing it restores the intended behavior. Existing tests already cover all platforms and edge cases.

## Summary by CodeRabbit

* **Bug Fixes**
  * Re-enabled multiprocessing on Linux. Worker count now scales with available CPU cores and GPU count, restoring parallel processing previously disabled on non-Windows platforms.

* **Tests**
  * Re-activated a unit test covering worker count behavior so the scaling logic is now validated in the test suite.


bittremieux (commented):
I'm not sure that your description is correct: the PyTorch issue refers to DataLoader issues on Apple Silicon, whereas the comment in our code refers to a deadlocking issue on Linux, not macOS.

The fact that the Ubuntu tests ran for 365 minutes before being cancelled also indicates that there's still an issue on Linux.

Can you look into this a bit more?

PewterZz:
Thanks for the thorough review. I investigated the actual root cause: it's a CUDA + fork deadlock on Linux (not related to pytorch/pytorch#70344). The fix requires using a non-fork multiprocessing context, but `depthcharge.preprocessing.set_mz_range()` returns a bound method that can't be pickled, blocking both `spawn` and `forkserver` contexts. The issue ultimately requires a fix in `depthcharge` or a significant refactor of casanovo's dataset class. Closing this PR and leaving `return 0` in place as the correct workaround.

PR #598, opened 2026-03-23
Fix overwrite protection using output_root_name instead of output_root

Hrishitcodethis:
Fix the overwrite check using the wrong variable

The overwrite protection was using `output_root` instead of
`output_root_name`. When `--output_root` isn’t passed,
`output_root` is None, so it ends up checking for `None.mztab`.

Updated both occurrences to use `output_root_name` so the check
runs against the actual resolved filename.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected output-file validation so existence is checked only when overwrite is disabled, and now aligns with the actual result path logic. This prevents incorrect skipping of processing and ensures results are produced and saved reliably during run and search operations.


bittremieux:
Something seems to have gone wrong and the diff is now empty. Could you push your changes again @Hrishitcodethis?

Hrishitcodethis:
yea will check, give me some time

Hrishitcodethis:
should i make the changes that the bots have suggested

bittremieux:
Yes.

Hrishitcodethis:
I have re-pushed

bittremieux:
For some reason I don't get any changes in the diff. 🤔
image

Hrishitcodethis:
I am checking it

Hrishitcodethis:
I think the issue is I forked from main and this pr is targeting dev branch

Hrishitcodethis:
Are the changes I am doing already in dev? Like is the issue still open or did someone else pushed same changes into dev

bittremieux:
No, this issue has not been addressed.

Hrishitcodethis:
Okay should i redo it with the base as dev this time, or should i change the current base to main because i had forked from main

bittremieux:
Yes, new changes are indeed first collected in `dev` rather than `main`.

bittremieux (commented):
Ok, the reason that this PR is empty because this has already been fixed on `dev` [several months ago](https://github.com/Noble-Lab/casanovo/blame/dev/casanovo/casanovo.py#L183).

PR #603, opened 2026-03-24
Addressed the duplicate prediction TODO asked in the code

mhmdkardosha:
**Title:** Fix: Retain highest score for duplicate peptide predictions in beam search

**Description:**
This PR resolves a `TODO` in the beam search decoding process (`_cache_finished_beams`) where duplicate peptide sequences were not being evaluated for their highest score, and it corrects stale documentation regarding how scores are stored in the cache. 

**Changes Made:**
* **Duplicate Prediction Handling:** Updated the caching logic in `model.py`. When a finished beam produces a peptide sequence that is already in the prediction cache, the model now compares their scores. If the newly predicted duplicate has a higher confidence score, it replaces the older entry and re-heapifies the priority queue to maintain bounds and order. 
* **Docstring Correction:** Fixed the `pred_cache` docstring to accurately reflect that the heap stores *positive* `peptide_score` values, removing the outdated reference to "(negated) peptide score".

**Motivation / Context:**
During beam search, different decoding paths can sometimes converge on the exact same final amino acid sequence but with different probability scores. Previously, duplicates were bypassed. This change ensures that if a spectrum produces the same peptide via multiple beams, the prediction with the highest possible confidence score is the one retained and reported to the user.

**Testing:**
* Verified that targeted unit tests (e.g., `test_duplicate_peptide_scores`, `test_cache_finished_beams`) pass successfully and enforce the new duplicate-updating behavior.
* Confirmed that the `pred_cache` heap maintains its maximum `n_beams` length.

This was made with the help of AI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved beam-search caching so identical candidate sequences are now compared and the higher-scoring prediction is preserved, preventing valid peptides from being discarded and improving final prediction quality.

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -563,7 +563,7 @@ class Spec2Pep(pl.LightningModule):
                     heapadd = heapq.heappush
                 else:
                     heapadd = heapq.heappushpop
-                
+
                 heapadd(pred_cache[spec_idx], peptide_entry)
 
     def _get_topk_beams(
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -502,7 +502,7 @@ class Spec2Pep(pl.LightningModule):
         ]
             Priority queue with finished beams for each spectrum,
             ordered by peptide score. For each finished beam, a tuple
-            with the (negated) peptide score, a random tie-breaking
+            with the peptide score, a random tie-breaking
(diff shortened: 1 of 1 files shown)


mhmdkardosha:
Hello @bittremieux. I hope you look at this.

bittremieux (approved):

PR #607, opened 2026-03-30
fix: removed spurious dir_okay=False check on --output_root (#568)

### LINKED ISSUE #568: Unneeded check for directory
## Describe the Issue

If I specify the same string for the `--output_dir` and the `--output_root`, casanovo issues the error `Invalid value for '-o' / '--output_root': File 'train-casa562-mssv' is a directory`.  This is incorrect behavior.  It should be perfectly OK to have these two strings by the same.  The directory name check, if it's done at all, should be carried out w.r.t. creation of individual files.

GauravSRC:
## Problem

Fixes #568.

When a user specifies the same string for `--output_dir` and
`--output_root` (e.g. both set to `train-casa562-mssv`), Casanovo
immediately exits with:

> Invalid value for '-o' / '--output_root': File 'train-casa562-mssv' is a directory.

This happens because `--output_root` was typed as
`click.Path(dir_okay=False)`, which instructs Click to reject any value
that resolves to an existing directory -- even before Casanovo runs.

## Root Cause

`--output_root` is purely a **base name** for output files
(`<root>.log`, `<root>.mztab`, etc.). It is not a directory path.
The directory-existence check in `click.Path(dir_okay=False)` is
therefore incorrect for this parameter. Real output-file conflict
detection is already handled by `utils.check_dir_file_exists()` inside
`_setup_output()`.

## Fix

Changed `type=click.Path(dir_okay=False)` → `type=str` for the
`--output_root` option in `_SharedFileIOParams` (`casanovo/casanovo.py`).

## Tests

Added `tests/unit_tests/test_casanovo.py` with a regression test
(`test_output_root_same_name_as_existing_dir`) that:
1. Creates a directory whose name equals the intended `--output_root` value.
2. Invokes `casanovo configure` with `--output_dir` and `--output_root`
   both set to that name.
3. Asserts the CLI accepts the input and exits cleanly (exit code 0).

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI option --output_root now accepts an existing directory name without triggering "is a directory" errors; invalid values (empty, "."/ "..", or path-like strings) are rejected with a clear error.
  * Log and warning messages consolidated to single-line formats for clearer CLI output.

* **Tests**
  * Added a regression test ensuring configure succeeds when --output_root points to an existing directory (with force overwrite).


bittremieux:
Already fixed on `dev` in 342e6cd.

PR #613, opened 2026-04-02
refactor: calculate shuffle buffer size as number of batches instead of total samples

### LINKED ISSUE #611: shuffle_buffer_size buffers batches, not samples
## Describe the Issue
shuffle_buffer_size in config.yaml is documented as:
> \# Number of samples to buffer while randomly shuffling the training data.

However, it actually controls the number of batches buffered by ShufflerIterDataPipe. This is because LanceDataset yields batches and ShufflerIterDataPipe treats each yielded item as one buffer entry, regardless of how many samples it contains.

This leads to buffering 10,000 batches instead of samples, delaying training startup.

## Expected Behavior
shuffle_buffer_size should either:                                                                                                                                                                                       

1.  Be converted to a number of batches internally (i.e. buffer_size = shuffle_buffer_size // train_batch_size), so users can reason about it in samples as documented, or
2. Be documented as the number of batches and set to a much smaller default.

mhmdkardosha:
## Problem

`shuffle_buffer_size` is documented in `config.yaml` as:

> #Number of samples to buffer while randomly shuffling the training data.

However, it is passed directly as `buffer_size` to `ShufflerIterDataPipe`, which treats each yielded item as one buffer entry. Since the underlying `LanceDataset` yields **batches** (not individual samples), the actual behavior is buffering `shuffle_buffer_size` batches.

With the default values (`shuffle_buffer_size=10,000`, `train_batch_size=32`), this means **320,000 samples** are buffered instead of the expected 10,000!

## Fix

Convert the user-facing sample count to a batch count before passing it to `ShufflerIterDataPipe`:

```python
buffer_batches = max(1, self.shuffle_buffer_size // self.train_batch_size)
```

This preserves the documented semantics — users continue to reason about `shuffle_buffer_size` in terms of samples — while the internal conversion ensures the correct number of samples are actually buffered. A `max(1, ...)` guard ensures the buffer is always at least 1 batch.

## What changes

- **`casanovo/denovo/dataloaders.py`**: In `_make_dataset`, divide `shuffle_buffer_size` by `train_batch_size` when constructing `ShufflerIterDataPipe`.

## Notes
Fixes #611 

## Release Notes

* **Bug Fixes**
  * Updated training data shuffling mechanism to dynamically scale buffer size based on batch size, reducing memory usage while maintaining proper data randomization during model training.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Noble-Lab/casanovo/pull/613?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -150,15 +150,6 @@ class Config:
         for key, val in self._config_types.items():
             self.validate_param(key, val)
 
-        if (
-            self._params["train_batch_size"] is not None
-            and self._params["train_batch_size"] <= 0
-        ):
-            raise ValueError(
-                f"train_batch_size must be a positive integer, "
-                f"got {self._params['train_batch_size']}."
-            )
-
         self._params["n_workers"] = utils.n_workers()
 
         if self._params["accelerator"] == "auto" and utils.is_apple_silicon():
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -150,6 +150,15 @@ class Config:
         for key, val in self._config_types.items():
             self.validate_param(key, val)
 
+        if (
+            self._params["train_batch_size"] is not None
+            and self._params["train_batch_size"] <= 0
+        ):
+            raise ValueError(
+                f"train_batch_size must be a positive integer, "
+                f"got {self._params['train_batch_size']}."
+            )
+
         self._params["n_workers"] = utils.n_workers()
 
         if self._params["accelerator"] == "auto" and utils.is_apple_silicon():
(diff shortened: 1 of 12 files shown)


casanovo/config.py
```diff
@@ -150,6 +150,15 @@ def __init__(self, config_file: Optional[str] = None):
         for key, val in self._config_types.items():
             self.validate_param(key, val)
 
+        if (
```
bittremieux: **todo:** Can you remove this part and the corresponding test? Config value validation will be handled more robustly in #590.

bittremieux (changes requested):
Looks good, a small request to keep the changes isolated to the `buffer_size` functionality only. Config validation is pending in another PR, so doesn't need to be added here.

bittremieux (approved):

PR #614, opened 2026-04-02
Fix shuffle buffer semantics for pre-batched training data

### LINKED ISSUE #611: shuffle_buffer_size buffers batches, not samples

maxprogrammer007:
## Problem

`shuffle_buffer_size` is documented as a number of samples, but the training pipeline passes pre-batched items to `ShufflerIterDataPipe`. That means the configured value was being interpreted as a number of batches, which made the effective shuffle window much larger than intended.

## Fix

This PR converts the sample-based `shuffle_buffer_size` into a batch count before passing it to the shuffler. The conversion is done in the data loader with a `max(1, ...)` guard so the buffer is always valid.

I also updated the config documentation to reflect the actual behavior and added a regression test to lock the semantics down.

## Why this is better than PR #613

- It keeps the user-facing meaning of `shuffle_buffer_size` unchanged: it still represents samples.
- It makes the conversion explicit and isolated in a helper, which is easier to read and maintain.
- It adds a regression test that verifies the exact buffer passed to `ShufflerIterDataPipe`.
- It avoids changing the config contract or relying on implicit behavior.

## Validation

- `pytest tests/unit_tests/test_unit.py -k 'shuffle_buffer_size_is_converted_to_batches or log_training_set_size_shuffled'`

Fixes #611 

## Summary by CodeRabbit

* **New Features**
  * Logs training, validation, and test dataset sizes during data module setup.
* **Tests**
  * Added unit tests that validate the emitted dataset-size logs and the shuffle-buffer conversion behavior.
* **Documentation**
  * Clarified config comments to explain that the shuffle buffer value (configured in samples) is converted internally to a batch count.


maxprogrammer007:
@bittremieux can this be merged ?

maxprogrammer007:
### Reply to: "Missing regression test for buffer size conversion"

Implemented in this PR. I added a dedicated regression test that monkeypatches `ShufflerIterDataPipe`, captures the `buffer_size` passed by `DeNovoDataModule`, and asserts the sample-to-batch conversion behavior directly in [tests/unit_tests/test_unit.py](tests/unit_tests/test_unit.py#L1964).

The production conversion is implemented in [casanovo/denovo/dataloaders.py](casanovo/denovo/dataloaders.py#L276) and used at [casanovo/denovo/dataloaders.py](casanovo/denovo/dataloaders.py#L271).

I intentionally use `ceil` plus `max(1, ...)` so the effective buffered samples are never below the user-configured sample target.

---

### Reply to: "Consider adding defensive check for n_spectra attribute"

Good suggestion, thanks. I am keeping this PR scoped to the shuffle buffer semantics fix from issue #611, so I am not changing `_get_n_spectra` behavior in this PR.

Today, the dataset types used in this path expose `n_spectra`, and CI is green across platforms. I can open a small follow-up PR to add defensive fallback logic for `_get_n_spectra` if you want that improvement separately.

mhmdkardosha:
Hi @bittremieux. I want to flag that this is a duplicate of my earlier PR (#613). The author was aware of my open PR while opening his own. I’d appreciate it if you could look into this.

bittremieux:
Thanks for the contribution @maxprogrammer007. However, typically we don't want to incentivize competing PRs, but rather that we can all learn from each other. For example, rather than opening a new PR, you can provide comments or a code review on an existing PR if you find further opportunities for improvement. I'll close this one in favor of #613.

PR #617, opened 2026-04-04
Implement cache expiration for URL-based weights

### LINKED ISSUE #616: _get_weights_from_url makes redundant network request on every invocation even when cache is valid
There is an issue in code practise in casanovo.py where request is redundant and making it slow and affecting the **performance**

In casanovo/casanovo.py, the `_get_weights_from_url `function performs a HEAD request to the remote URL on every single run, even when a valid cached file already exists. This adds unnecessary network latency to every invocation.
Additionally, `url_last_modified `is initialized to `0` as a fallback. This means if the HEAD request returns a non-ok status, the cache staleness check becomes:
`url_last_modified = 0`

# This is almost always True, so cache is used   but only by accident
`_if cache_time.st_mtime > url_last_modified:`
The cache is used correctly in this case, but only coincidentally. If the fallback value were ever changed, this logic would silently break and trigger unnecessary re-downloads on every run._
**Steps To Reproduce**

- Run casanovo sequence with a URL-based model for the first time to populate the cache:
- `casanovo sequence --model https://example.com/model.ckpt myfile.mzML`

- Run the exact same command again immediately after
- Observe that a HEAD request is still made to the remote URL despite the cache being fresh

**Expected Behavior**
If a cached file exists and was downloaded recently, Casanovo should skip the network request entirely and use the cached file without any remote calls. A configurable cache TTL (e.g. 24 hours) should determine when a freshness check is actually needed.
# Suggested fix 
in casanovo/casanovo.py:

Image

```
CACHE_TTL_SECONDS = 86400  # 24 hours

if cache_file_path.is_file() and not force_download:
    cache_age = time.time() - cache_file_path.stat().st_mtime
    if cache_age < CACHE_TTL_SECONDS:
        logger.info(
            "Model weights %s retrieved from local cache", file_url
        )
        return cache_file_path
    # Only make HEAD request if cache is older than TTL
    ...
```

Additional Context
This issue affects every user running casanovo sequence or casanovo db_search with a URL-based --model argument. On slow or unreliable network connections, this redundant request adds noticeable latency and introduces a failure point on every run even when the cache is perfectly valid.

codewithEshaYoutube:
Added CACHE_TTL_SECONDS constant to define cache expiration time for URL-based weights. Updated cache handling logic to utilize the new constant for checking freshness before making network requests.

## Summary by CodeRabbit

* **New Features**
  * Time-based cache freshness: recent cached weight files are used without re-downloading, reducing network calls and speeding startup.

* **Bug Fixes**
  * Better handling of remote metadata with safe fallbacks to preserve usable cached copies.
  * Improved network timeouts for metadata checks and downloads to avoid hangs.

* **Tests**
  * Added a unit test verifying fresh cache entries bypass network requests.


bittremieux:
Please make sure that CodeRabbit comments have been addressed and all tests pass, then @ me to do a code review.

codewithEshaYoutube:
Fixed in commit bb6556e — url_last_modified is now initialized to float("-inf")

codewithEshaYoutube:
Fixed in commit bb6556e — requests.get now has timeout=60
Fixed in commit bb6556e — parsedate_to_datetime is now wrapped in except (TypeError, ValueError)

codewithEshaYoutube:
@bittremieux  All CodeRabbit comments have been consistently addressed and a test case has been added to cover the cache expiration logic.
Checks are passing . Its ready for your review!

bittremieux:
@codewithEshaYoutube can you make sure that all tests and the linter pass? Please tag me for a code review once this is the case.

bittremieux:
Hi @codewithEshaYoutube is this PR still active? The first step is making sure that the linting and unit test checks all pass.

bittremieux:
Stale PR that doesn't follow the contributing guidelines, closing.

PR #628, opened 2026-04-27
fix: allow val_check_interval to be a float in (0, 1] or an int >= 1

### LINKED ISSUE #627: allow val_check_interval to be a float
In pytorch, val_check_interval can be either a float in the range [0,1] or an int.  If it's a float, it specifies how frequently the validation set is analyzed per epoch, whereas if it's an int it specifies the same thing in number of batches (steps).  Currently, casanovo only allows it to be an int. I think we should allow either, and document it as such in the config file.  Furthermore, I think the default value should be 1.0.

Ghraven:
Closes #627

## What

Adds `_parse_val_check_interval()` — a dedicated validator that accepts PyTorch Lightning's full range for `val_check_interval`:

- **`float` in `(0.0, 1.0]`** — fraction of an epoch between validations
- **`int >= 1`** — number of training batches between validations

Previously the `_config_types` dict mapped `val_check_interval` to `int`, so passing `0.25` or `1.0` would be silently coerced or raise a confusing error.

## Changes

- `casanovo/config.py`: replace `val_check_interval=int` with `val_check_interval=_parse_val_check_interval`
- New module-level function `_parse_val_check_interval` with clear docstring and `ValueError` for out-of-range values
- Default value in `config.yaml` (`50_000`) continues to work as an int

## Example

```yaml
# validate every quarter of an epoch
val_check_interval: 0.25

# validate every 10 000 steps (existing behaviour)
val_check_interval: 10000
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced validation of the validation check interval configuration parameter to properly handle different input types and provide clearer error messages when invalid values are supplied.

### DIFF
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -28,11 +28,15 @@ _config_deprecated = dict(
 
 
 def _parse_val_check_interval(value) -> Union[int, float]:
-    """Parse val_check_interval as either a float (0, 1] or a positive int.
+    """Parse val_check_interval as either an int >= 1 or a float in (0, 1].
 
     PyTorch Lightning accepts:
-    - float in (0.0, 1.0]: fraction of epoch between validations
     - int >= 1: number of training batches between validations
+    - float in (0.0, 1.0]: fraction of epoch between validations
+
+    The original type is preserved so PyTorch Lightning's strict type
+    semantics are respected: ``int(1)`` stays ``int`` (one-batch interval),
+    while ``float(1.0)`` stays ``float`` (end-of-epoch fraction).
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


Lilferrit:
I would hold off on implementing this until #545 is implemented (for which there is a PR open already, #590). The implementation of this will change once we switch over to omgaconf for handling the config parsing.

Ghraven:
Thanks for the context, @Lilferrit — I was not aware of #590 or the planned omegaconf migration in #545. I will hold off on this until that work lands. Happy to revisit and rebase on top of the new config structure once #590 is merged if that would still be useful.

Ghraven:
Hi! Just a gentle check-in on this one. If you'd like any changes to the validation logic or the approach isn't quite right for the project, I'm happy to revise. No pressure at all — thanks so much for maintaining this!

bittremieux:

PR #649, opened 2026-06-04
fix: bound model weights download requests

Ghraven:
## Problem

URL-based model weight resolution uses `requests.head()` for cache freshness checks and `requests.get()` for streaming downloads without a timeout. If the remote server accepts the connection but stalls, Casanovo can wait indefinitely while resolving weights.

## Before / After

Before, both the HEAD freshness check and the streaming download could block forever on a slow or unresponsive server.

After, both requests use the same 30 second timeout constant, so network stalls fail in a bounded way and the existing cache fallback behavior for HEAD failures still applies.

## Verification

- `python -m py_compile casanovo\casanovo.py tests\unit_tests\test_unit.py`
- Attempted `python -m pytest tests/unit_tests/test_unit.py::test_get_weights_from_url`, but the local environment is missing `psims` from `tests/conftest.py`, so pytest could not start.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model weight download reliability by adding time limits to remote requests used for downloading and cache checking. This prevents indefinite hangs during weight retrieval and cache validation, making model loading more robust and responsive for end users.

### DIFF
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -954,9 +954,7 @@ def _get_weights_from_url(
         url_last_modified = 0
 
         try:
-            file_response = requests.head(
-                file_url, timeout=REQUEST_TIMEOUT
-            )
+            file_response = requests.head(file_url, timeout=REQUEST_TIMEOUT)
             if file_response.ok:
                 if "Last-Modified" in file_response.headers:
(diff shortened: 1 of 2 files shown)


bittremieux:

PR #651, opened 2026-06-07
perf: vectorize beam search decoding operations

GauravSRC:
## Summary

This PR implements the vectorization and tensor-native improvements from
commit haleelsada/casanovo_cpu@37d9ea2, excluding int8 quantization.

Two changes from the original commit (cumulative_masses loop and
_finish_beams tokenizer loop) were superseded by refactoring already on
main and are not included.

## Changes

**1. `_cache_finished_beams` -- deferred CPU transfer for AA scores**
AA scores are kept on GPU as a tensor until they need to be written to
the prediction cache, removing a `.cpu().numpy()` round-trip that
previously occurred once per finished beam per decode step.

**2. `_peptide_score` -- native tensor support**
`_peptide_score` now accepts `torch.Tensor` inputs in addition to
`np.ndarray`. The numpy path is fully preserved for database search
and all existing callers. The tensor path eliminates the GPU→CPU
synchronization in the de novo decoding path. The `fits_precursor_mz`
parameter is retained so out-of-tolerance beams correctly receive the
-1 score penalty. Mixed numpy/tensor inputs in the batched path are
normalised to tensors before indexing.

**3. `compile_model()` -- optional torch.compile wrapper**
Adds an opt-in `compile_model()` method wrapping the encoder and
decoder with `torch.compile(dynamic=True)`. `dynamic=True` is required
for beam search because the token sequence grows by one element each
decode step. Not called automatically; users call it after loading
weights if they want the additional throughput.

**4. Restores `fits_precursor_mz` in `_peptide_score`**
The parameter was inadvertently removed from main. This PR restores it
so the test suite passes and out-of-tolerance scoring is correct.

## Timing results

Measured on NVIDIA L4 (23.6 GB VRAM), PyTorch 2.7.1+cu128,
`multi-enzyme-simple.test.mgf`, 5000 spectra per measurement.

| Batch size | Baseline (spec/s) | All patches (spec/s) | Speedup |
|---|---|---|---|
| 8  | 14.1 | 14.7 | 1.04x |
| 32 | 43.1 | 46.4 | 1.08x |
| 128 | 81.5 | 91.2 | 1.12x |
| 512 | 97.5 | 110.7 | 1.14x |

Speedup grows with batch size because the patches reduce python-level
overhead that becomes a larger fraction of per-step cost as GPU
utilisation increases. At bs=1 the benefit is ~1% (97% of runtime is
sequential AR decoding). The benefit is most visible at batch sizes ≥ 32.

## Notes

`test_spectrum_id_mgf` fails on the `main` branch before this PR
(scan ID format mismatch from a depthcharge version update, unrelated
to these changes). This PR does not touch any data loading code.

**Codecov:** Patch coverage is below the 95.47% target because the new 
torch.Tensor path in `_peptide_score` and the `compile_model()` method 
are not exercised by the existing test suite (tests run on CPU without 
torch.compile). The numpy path and all existing code paths remain fully 
covered. All lint and build/test checks pass.

## Summary by CodeRabbit

- **Performance Improvements**
  - Reduced CPU–GPU transfers during beam-search scoring by performing amino-acid and peptide scoring on the GPU until post-processing.
  - Improved handling when the terminal stop marker is missing to ensure scoring remains correct.

- **Refactor**
  - Updated peptide scoring to support both single and batched inputs, returning results in the same data type as the inputs and using provided lengths for accurate endpoint evaluation.


casanovo/denovo/model.py
```diff
@@ -1493,39 +1533,81 @@ def get_lr_factor(self, epoch):
 
 
 def _peptide_score(
-    aa_scores: np.ndarray,
-    lengths: Optional[np.ndarray] = None,
-) -> Union[float, np.ndarray]:
+    aa_scores: Union[np.ndarray, torch.Tensor],
+    lengths: Optional[Union[np.ndarray, torch.Tensor]] = None,
+    fits_precursor_mz: Union[bool, np.ndarray, torch.Tensor] = True,
```
bittremieux: **todo:** The precursor mass filter was deliberately rather than inadvertently removed. Please revert this change.

casanovo/denovo/model.py
```diff
@@ -1493,39 +1533,81 @@ def get_lr_factor(self, epoch):
...
         An array of peptide lengths, required when `aa_scores` is a 2D
         (batched) array.
 
     Returns
     -------
-    peptide_score : float or np.ndarray
+    peptide_score : float, np.ndarray, or torch.Tensor
         The calculated peptide score or an array of scores for the batch.
     """
+    if isinstance(aa_scores, torch.Tensor):
```
bittremieux: **todo:** This near-duplication between the Tensor and the NumPy path is not ideal. Please try to consolidate it into a single code path. I.e. it might be possible to convert NumPy arrays to Tensors without actually copying the data (but find the best place to do this).

casanovo/denovo/model.py
```diff
@@ -1537,7 +1619,11 @@ def _peptide_score(
         cumsum = np.cumsum(log_scores, axis=1)
         batch_size = aa_scores.shape[0]
         idx = np.arange(batch_size)
-        peptide_log_scores = cumsum[idx, np.maximum(lengths - 1, 0)]
-        peptide_scores = np.exp(peptide_log_scores)
+        peptide_scores = np.exp(cumsum[idx, np.maximum(lengths - 1, 0)])
 
+        if isinstance(fits_precursor_mz, (bool, np.bool_)):
```
bittremieux: **todo:** As mentioned above, this is incorrect (also line 1610).

casanovo/denovo/model.py
```diff
@@ -200,6 +200,38 @@ def __init__(
...
+            idx = self.tokenizer.index.get(aa)
+            if idx is not None:
+                self.token_masses[idx] = mass
+
+    def compile_model(self) -> None:
+        """
+        Apply ``torch.compile`` to the encoder and decoder for
+        improved throughput, particularly at larger batch sizes.
+
+        This is an optional, opt-in optimization. Call after loading
```
bittremieux: **question:** Why is this optional? Is there a downside to always compiling?

Also, now this method is not called anywhere, and there is no way to enable it. So it's basically some dead code?

casanovo/denovo/model.py
```diff
@@ -200,6 +200,38 @@ def __init__(
             persistent=False,
         )
 
+        # Register tensor buffer for amino acid token masses
+        self.register_buffer(
```
bittremieux: **question:** Where is this buffer actually used?

bittremieux (changes requested):
Please see the comments. Also make sure that all PRs are opened against `dev`, not `main`.

casanovo/denovo/model.py
GauravSRC: got it,reverting it now.

casanovo/denovo/model.py
GauravSRC: This buffer was introduced for an earlier version of the cumulative masses vectorization (P1) that was superseded by main's refactoring. Since nothing in the current codebase references it, removing it now.

casanovo/denovo/model.py
GauravSRC: You are right-I have removed from this pr. The main downside of always compiling is significant startup latency (first-call compilation takes 30–60s), which makes it unsuitable as an always-on default. Happy to revisit this as a separate PR with an opt-in config flag if it's worth pursuing.

casanovo/denovo/model.py
```diff
@@ -1493,51 +1501,58 @@ def get_lr_factor(self, epoch):
...
+    # contiguous CPU arrays), enabling a single unified computation path.
+    if return_numpy:
+        aa_scores = torch.as_tensor(aa_scores)
 
-    # BATCH PATH: database search
-    else:
-        if lengths is None:
-            raise ValueError("`lengths` must be provided for batched input.")
+    eps = torch.finfo(torch.float64).eps
+    log_scores = torch.log(torch.clamp(aa_scores.double(), eps, 1))
```
bittremieux: **question:** Is the conversion to `double` essential?

GauravSRC: Yes--not essential for correctness, but it helps avoid float32 rounding error accumulating over up to 100 decoding steps and keeps precision consistent with the epsilon clamp and the original numpy implementation (which also used float64).

bittremieux: Did you observe a float32 rounding error or is that an assumption? I don't see an explicit conversion to float32 in the original NumPy code; what is the type of `aa_scores` when it comes in?

GauravSRC: that was an assumption, not something I actually observed empirically.Looking at the code, aa_scores comes from softmax(decoder_output), and the decoder/scores tensor defaults to float32 (no explicit dtype is set), so it is  float32 going into _peptide_score. I don't have measured evidence that this causes a real precision issue in practice. I will drop the explicit .double() cast if you'd prefer--let me know.

bittremieux: Yes, unless there's evidence that it's needed it can be removed.

casanovo/denovo/model.py
```diff
@@ -1493,51 +1501,58 @@ def get_lr_factor(self, epoch):
...
-        return peptide_scores
+    if aa_scores.ndim == 1:
+        # FAST PATH: de novo inference — single peptide.
+        peptide_score = torch.exp(torch.sum(log_scores))
+        return peptide_score.item() if return_numpy else peptide_score
+
+    # BATCH PATH: database search — padded batch of peptides.
+    if lengths is None:
+        raise ValueError("`lengths` must be provided for batched input.")
+    if not isinstance(lengths, torch.Tensor):
```
bittremieux, line 1550: **todo:** Please address the comment here: https://github.com/Noble-Lab/casanovo/pull/651#pullrequestreview-4538937313

bittremieux (changes requested):

bittremieux (approved):

PR #653, opened 2026-06-11
fix: skip fresh cached URL weight checks

### LINKED ISSUE #616: _get_weights_from_url makes redundant network request on every invocation even when cache is valid

Ghraven:
## Problem

Closes #616.

`_get_weights_from_url()` made a `HEAD` request every time a URL-backed model file already existed in the cache. That meant repeat invocations still depended on network latency and remote availability even when the local cache had just been populated.

## Before / after

Before:
- First run downloaded and cached the URL model weights.
- A second immediate run reused the cached file, but still performed a remote `HEAD` request first.

After:
- Cached URL weights newer than the 24-hour TTL are returned immediately without a remote `HEAD` request.
- Stale cached files still use the existing `HEAD` / `Last-Modified` freshness check before deciding whether to re-download.
- `force_download=True` still bypasses the cache.

## Verification

- `python -m pytest tests/unit_tests/test_unit.py -k get_weights_from_url`
- `python -m ruff check --extend-ignore E402 casanovo/casanovo.py tests/unit_tests/test_unit.py`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Optimized weight file caching to validate locally cached copies against a time-based validity window, eliminating unnecessary remote verification requests for recently cached weight files and improving load performance.

* **Tests**
  * Enhanced unit tests to verify cache validation logic and confirm that remote verification requests are only made when the cached file is no longer within the acceptable time window.

### DIFF
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -47,6 +47,8 @@ click.rich_click.USE_MARKDOWN = True
 click.rich_click.STYLE_HELPTEXT = ""
 click.rich_click.SHOW_ARGUMENTS = True
 
+_URL_WEIGHTS_CACHE_TTL_SECONDS = 24 * 60 * 60
+
 
 class _SharedFileIOParams(click.RichCommand):
     """File IO options shared between most Casanovo commands"""
@@ -950,6 +952,12 @@ def _get_weights_from_url(
 
     if cache_file_path.is_file() and not force_download:
         cache_time = cache_file_path.stat()
+        if time.time() - cache_time.st_mtime < _URL_WEIGHTS_CACHE_TTL_SECONDS:
+            logger.info(
+                "Model weights %s retrieved from local cache", file_url
+            )
+            return cache_file_path
+
         url_last_modified = 0
 
         try:
--- a/tests/unit_tests/test_unit.py
+++ b/tests/unit_tests/test_unit.py
@@ -444,8 +444,10 @@ class MockResponseHead:
         self.last_modified = None
         self.is_ok = True
         self.fail = False
+        self.request_counter = 0
 
     def __call__(self, url):
+        self.request_counter += 1
         if self.fail:
             raise requests.ConnectionError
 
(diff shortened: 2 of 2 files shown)


bittremieux:

PR #669, opened 2026-07-12
Docstring changes

bhavyak7-lab:
Changes made: 
- Moved the ckpt regex to the top of the file 
- Made docstring more detailed for model loading

## Summary by CodeRabbit

* **Documentation**
  * Clarified model configuration guidance, including supported model identifiers and behavior when no model, URL, or file is provided.

### DIFF
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -31,6 +31,14 @@ warnings.filterwarnings(
     ".*Converting mask without torch.bool dtype to bool*",
 )
 
+# The regex pattern below describes the the model weight
+# naming pattern.
+_CKPT_RE = re.compile(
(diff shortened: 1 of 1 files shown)


casanovo/casanovo.py
```diff
@@ -660,11 +659,12 @@ def setup_model(
     Parameters
     ----------
     model : str | None
-        May be a file system path, a URL pointing to a .ckpt file, or
-        None. If `model` is a URL the weights will be downloaded and
```
bittremieux, line 664: **question:** What does "selector" mean? The type hint also indicates `str` or `None`, so I don't know what you mean.

bittremieux (changes requested):

casanovo/casanovo.py
bhavyak7-lab: The selector is supposed to refer to the type of model (timstof, orbitrap, etc.). I have updated the wording to make that clearer.

PR #674, opened 2026-07-29
Stop writing Lance datasets and mzTab output into the working directory during testing

BioGeek:
When you run the full test suite in a local checkout, a lot of unchecked directories and files are left in the root directory. These test artifiacts are directories named after the tests that created them, each holding Lance datasets with random UUID filenames, plus a stray test.mztab.

The cause was that eight call sites passed `tmp_path.name` where a path was expected. `.name` discards the parent, so a pytest temporary directory such as `/tmp/pytest-of-user/pytest-9/test_log_training_set_size0` collapses to the bare relative name `test_log_training_set_size0`, which is then created in whatever the working directory happens to be.

A ninth site, `test_db_search_no_model_raises `, passed the literal `"test.mztab"`, so its output landed in the current directory too. That test did not take `tmp_path`, so the fixture is added to its signature.

Fixed by passing `str(tmp_path)`, which is what the surrounding code expects. No test logic or assertion changes.

Because the UUID filenames cannot be matched by a fixed pattern, a `.gitignore  rule would have needed a directory glob broad enough to risk masking real files. Putting the output where pytest already manages and cleans it is the narrower fix.

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by using temporary filesystem paths for generated output.
  * Updated dataset-related tests to use temporary directories consistently.
  * Enhanced coverage for spectrum handling, preprocessing, logging, and validation scenarios.

### DIFF
--- a/tests/unit_tests/test_runner.py
+++ b/tests/unit_tests/test_runner.py
@@ -116,7 +116,9 @@ def test_db_search_no_model_raises(
         ),
     ):
         runner.db_search(
-            (str(mgf_small),), str(tiny_fasta_file), str(tmp_path / "test.mztab")
+            (str(mgf_small),),
+            str(tiny_fasta_file),
+            str(tmp_path / "test.mztab"),
         )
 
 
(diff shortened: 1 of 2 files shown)


BioGeek:
Ready for review @bittremieux

bittremieux (approved):
Thanks! 🙏

PR #677, opened 2026-08-09
fix: bound model weight HTTP requests

Ghraven:
## Problem

Remote model-weight resolution uses `requests.head()` to check cached file freshness and `requests.get(..., stream=True)` to download weights. Neither request had a timeout, so a slow or stalled remote endpoint could leave Casanovo waiting indefinitely while resolving model weights.

## Before / after behavior

Before, both the cache freshness request and the streamed download could block without a client-side bound.

After, both requests use the same 30 second timeout constant. The existing cached-file fallback for HEAD request failures is preserved, and failed downloads still surface through `raise_for_status()` / requests exceptions.

## Verification

- `python -m pytest tests/unit_tests/test_unit.py -k get_weights_from_url` - 1 passed, 117 deselected
- `python -m black --check casanovo/casanovo.py tests/unit_tests/test_unit.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a 30-second timeout to model-weight validation and downloads.
  * Improved reliability by preventing requests from hanging indefinitely.

* **Tests**
  * Added coverage confirming timeouts are applied during initial, cached, and forced downloads.

### DIFF
--- a/casanovo/casanovo.py
+++ b/casanovo/casanovo.py
@@ -47,6 +47,8 @@ click.rich_click.USE_MARKDOWN = True
 click.rich_click.STYLE_HELPTEXT = ""
 click.rich_click.SHOW_ARGUMENTS = True
 
+_MODEL_WEIGHT_REQUEST_TIMEOUT = 30
+
 
 class _SharedFileIOParams(click.RichCommand):
     """File IO options shared between most Casanovo commands"""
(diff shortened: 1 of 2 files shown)


bittremieux (approved):

Ghraven:
Thank you for the review and merge. I appreciate it.

PR #681, opened 2026-08-11
fix(config): handle empty user config files

Ghraven:
## Problem

When a user-provided YAML config file is empty, `yaml.safe_load()` returns `None`. Casanovo then tries to check deprecated option names with `old in self._user_config`, which raises a raw `TypeError: argument of type 'NoneType' is not iterable` before the existing config validation can explain what is wrong.

## Before / after behavior

Before, an empty user config failed with a low-level `TypeError`.

After, empty YAML is treated as an empty config mapping, so the existing missing-option validation raises the same clear `KeyError` path used for incomplete config files.

## Verification

- `python -m pytest tests/unit_tests/test_config.py` - 3 passed
- `python -m black --check casanovo/config.py tests/unit_tests/test_config.py`
- `python -m compileall casanovo/config.py tests\unit_tests\test_config.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Empty or null configuration files are now handled safely.
  * Clear errors are reported when required configuration options are missing.
  * Invalid non-mapping configuration values, such as lists, now produce a descriptive error.

* **Tests**
  * Added coverage for empty YAML values, null configurations, and invalid list-based configuration files.

### DIFF
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -115,7 +115,13 @@ class Config:
             self._user_config = {}
         else:
             with Path(config_file).open() as f_in:
-                self._user_config = yaml.safe_load(f_in) or {}
+                self._user_config = yaml.safe_load(f_in)
+                if self._user_config is None:
+                    self._user_config = {}
+                elif not isinstance(self._user_config, dict):
+                    raise TypeError(
+                        "Configuration file must define a mapping of options"
+                    )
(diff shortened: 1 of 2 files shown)


bittremieux (approved):

Ghraven:
Thank you for the review and merge. I appreciate it.

PR #682, opened 2026-08-12
Allow val_check_interval to be a float

### LINKED ISSUE #627: allow val_check_interval to be a float

ssiddhantsharma:
Closes #627.

`val_check_interval` now accepts a float in `[0, 1]` (fraction of a training
epoch) as well as an int (number of steps). The default changes to `1.0`
(validate once per epoch).

Lightning requires `check_val_every_n_epoch` to be set when `val_check_interval`
is a float, so it is `1` for floats and stays `None` for ints (unchanged
step-based behavior).

Notes:
- Behavior change: default validation cadence goes from every 50,000 steps to
  once per epoch.
- A float `0.0` follows Lightning and validates every step; happy to forbid it
  if you'd prefer.
- Left `config_timstof.yaml` at `50_000`; can align it to `1.0` if you want.

Added tests for the config coercion and the trainer setup.

## Summary by CodeRabbit

- **New Features**
  - Configure validation and checkpoint frequency using positive integer training-step counts or fractional epoch intervals from `0` to `1`.
  - Validation now runs once per epoch by default with an interval of `1.0`.

- **Bug Fixes**
  - Improved scheduling for fractional validation intervals while preserving step-based scheduling.
  - Added validation for unsupported interval values, including zero, out-of-range fractions, and booleans.

- **Documentation**
  - Updated FAQs and checkpoint guidance to explain validation interval options, defaults, and evaluation-based checkpoint timing.

### DIFF
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -26,6 +26,16 @@ _config_deprecated = dict(
 )
 
 
+def _int_or_float(value: Union[int, float]) -> Union[int, float]:
+    """Cast to an int (steps) or a float in [0, 1] (fraction of an epoch)."""
+    if isinstance(value, int):
+        return value
+    value = float(value)
+    if not 0.0 <= value <= 1.0:
+        raise ValueError("a float value must be in [0, 1]")
+    return value
+
+
 class Config:
     """
     The Casanovo configuration options.
@@ -70,7 +80,7 @@ class Config:
         log_metrics=bool,
         log_every_n_steps=int,
         lance_dir=str,
-        val_check_interval=int,
+        val_check_interval=_int_or_float,
         min_peaks=int,
         max_peaks=int,
         min_mz=float,
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -29,7 +29,7 @@ _config_deprecated = dict(
 def _int_or_float(value: Union[int, float]) -> Union[int, float]:
     """Cast to an int (steps) or a float in [0, 1] (fraction of an epoch)."""
     if isinstance(value, bool):
-        raise ValueError("must be an int or a float, not a bool")
+        raise TypeError("must be an int or a float, not a bool")
     if isinstance(value, int):
         if value < 1:
             raise ValueError("an int value must be at least 1")
--- a/casanovo/config.py
+++ b/casanovo/config.py
@@ -27,16 +27,32 @@ _config_deprecated = dict(
 
 
 def _int_or_float(value: Union[int, float]) -> Union[int, float]:
-    """Cast to an int (steps) or a float in [0, 1] (fraction of an epoch)."""
+    """Cast the validation interval to an int (steps) or float in [0, 1]."""
     if isinstance(value, bool):
-        raise TypeError("must be an int or a float, not a bool")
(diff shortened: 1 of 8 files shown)


ssiddhantsharma:
Kept the default at `50_000` steps rather than `1.0`. With `1.0`, validation runs every epoch, which the macos-15 CI runner (Apple Silicon/MPS) can't do: it hits the unimplemented `aten::_nested_tensor_from_mask_left_aligned` MPS op during validation (the same class of error the FAQ documents needing `PYTORCH_ENABLE_MPS_FALLBACK=1`). The float capability is fully intact, so `1.0` works whenever that fallback is set. Happy to switch the default to `1.0` if you'd like CI to set the fallback, or to handle the MPS path separately.

casanovo/config.py
```diff
@@ -26,6 +26,20 @@
 )
 
 
+def _int_or_float(value: Union[int, float]) -> Union[int, float]:
+    """Cast to an int (steps) or a float in [0, 1] (fraction of an epoch)."""
+    if isinstance(value, bool):
+        raise TypeError("must be an int or a float, not a bool")
```
bittremieux: **todo:** I recommend a more specific error message, accompanied by logging. Now the error message only talks about floats and ints, rather than specifying that this concerns the validation interval.

docs/file_formats.md
```diff
@@ -303,7 +303,7 @@ Similarly, in Casanovo evaluation mode only annotated MGF files are supported.
 
 <!-- TODO: when index files can be reused, document this here -->
 
-During training, Casanovo will save **checkpoint files** at every `val_check_interval` steps, specified in the configuration.
+During training, Casanovo will save a **checkpoint file** at the end of each training epoch, and update the best-validation checkpoint at the `val_check_interval` frequency specified in the configuration.
```
bittremieux: **todo:** I think that this should be changed, and that a checkpoint should be created every time the validation performance is evaluated.

bittremieux (changes requested):
Thanks for the contribution @ssiddhantsharma, and for being thorough with also updating the tests and documentation. I have a few small comments.

docs/file_formats.md
ssiddhantsharma: Hi, this is now reworded to say a checkpoint is saved every time the validation performance is evaluated, at the `val_check_interval` frequency.

casanovo/config.py
ssiddhantsharma: Thanks for the review. I addressed both in 0393902, the error messages now name `val_check_interval` and state the two accepted forms (a positive integer number of training steps, or a float in `[0, 1]` for a fraction of an epoch), and each is logged with `logger.error` before raising.

bittremieux (approved):
Thanks for the contribution and the updates, this looks good.

PR #688, opened 2026-08-21
Count and log spectra that receive no prediction

### LINKED ISSUE #660: Improving logging wrt beam search failures
The Casanovo log file reports when spectra get skipped because they have too few peaks or out-of-range charge state.  However, some spectra fail to receive a prediction because the beam search terminates without a valid peptide because the predicted sequence is shorter than the minimum peptide length, an invalid N-terminal modification was predicted, or a padding token was predicted.  We should keep a count of these and report that count in the log file.

Attached is a sample MGF file that contains 10 spectra but for which Casanovo only makes one prediction.

[missing-predictions.zip](https://github.com/user-attachments/files/29700376/missing-predictions.zip)

ssiddhantsharma:
Closes #660.

Some spectra receive no prediction because beam search terminates without a valid peptide (e.g. the sequence is shorter than the minimum length, an invalid N-terminal modification, or a padding token). These were silently dropped, so the number of reported PSMs could be less than the number of input spectra with no indication why.

The number of such spectra is now counted in `predict_step` (when a spectrum yields no beam), reset at the start of prediction, and logged at the end of a run, so every spectrum is accounted for as either scored or skipped.

Added a test that mocks a mix of predicted and empty beam results and asserts the count and its reset.

## Summary by CodeRabbit

* **Bug Fixes**
  * Spectra without valid peptide predictions are now skipped instead of producing invalid results.
  * Sequencing and database-search reports now count and display spectra lacking valid predictions.
  * Prediction statistics reset correctly at the start of each run and aggregate accurately across processing devices.
  * Runs with no missing predictions no longer produce unnecessary warnings, while missing-prediction counts are clearly logged.

### DIFF
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -163,6 +163,8 @@ class Spec2Pep(pl.LightningModule):
         self.calculate_precision = calculate_precision
         self.n_log = n_log
         self._history = []
+        # Count of spectra for which beam search returned no valid peptide.
+        self.n_missing_predictions = 0
         # Per-file validation metadata; set by ModelRunner.train() before fit.
         self.val_stems: list = []
         self.n_main_loaders: int = 0
@@ -969,6 +971,9 @@ class Spec2Pep(pl.LightningModule):
             batch["precursor_mz"],
             self.forward(batch),
         ):
+            if not spectrum_preds:
+                self.n_missing_predictions += 1
+                continue
             for peptide_score, aa_scores, peptide in spectrum_preds:
                 predictions.append(
                     psm.PepSpecMatch(
@@ -1026,6 +1031,19 @@ class Spec2Pep(pl.LightningModule):
                 )
         self._history.append(metrics)
         self._log_history()
+
+    def on_predict_start(self) -> None:
+        """Reset the count of spectra without a prediction."""
+        self.n_missing_predictions = 0
+
+    def on_predict_epoch_end(self) -> None:
+        """Log the number of spectra that did not receive a prediction."""
+        if self.n_missing_predictions > 0:
+            logger.info(
+                "%d spectra did not receive a prediction because beam "
+                "search did not return a valid peptide",
+                self.n_missing_predictions,
+            )
 
     def on_predict_batch_end(
         self, outputs: List[psm.PepSpecMatch], *args
--- a/casanovo/denovo/model.py
+++ b/casanovo/denovo/model.py
@@ -1037,18 +1037,12 @@ class Spec2Pep(pl.LightningModule):
         self.n_missing_predictions = 0
 
     def on_predict_epoch_end(self) -> None:
-        """Log the number of spectra that did not receive a prediction."""
-        n_missing = int(
+        """Aggregate the missing-prediction count across devices."""
+        self.n_missing_predictions = int(
             self.all_gather(
                 torch.tensor(self.n_missing_predictions, device=self.device)
             ).sum()
         )
-        if n_missing > 0 and self.trainer.is_global_zero:
-            logger.info(
-                "%d spectra did not receive a prediction because beam "
-                "search did not return a valid peptide",
-                n_missing,
-            )
 
     def on_predict_batch_end(
         self, outputs: List[psm.PepSpecMatch], *args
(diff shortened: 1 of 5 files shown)


casanovo/denovo/model.py
```diff
@@ -1027,6 +1032,24 @@ def on_validation_epoch_end(self) -> None:
...
+        self.n_missing_predictions = 0
+
+    def on_predict_epoch_end(self) -> None:
+        """Log the number of spectra that did not receive a prediction."""
+        n_missing = int(
+            self.all_gather(
+                torch.tensor(self.n_missing_predictions, device=self.device)
+            ).sum()
+        )
+        if n_missing > 0 and self.trainer.is_global_zero:
```
bittremieux: **todo:** This information should be passed to the [reporting function](https://github.com/Noble-Lab/casanovo/blob/main/casanovo/utils.py#L177) to synchronize end-of-run reporting in a consistent fashion.

Bonus points if we can harmonize logging of spectra skipped during preprocessing there as well.

bittremieux (changes requested):

casanovo/denovo/model.py
ssiddhantsharma: Done in 6ac34b6. The missing-prediction count is now aggregated across devices in `on_predict_epoch_end` and passed to `log_annotate_report` it's reported as part of the end-of-run report now.

On the bonus, that count currently only surfaces as a warning from depthcharge's parser, which Casanovo doesn't capture as a value. I'd be glad to follow up on surfacing it through the same report in a separate change.

bittremieux: Sounds good. It seems that there are a few new CodeRabbit comments. Can you resolve those and then tag me for a re-review afterwards? Thanks.

ssiddhantsharma: Thanks! Addressed the CodeRabbit comments in 041d96b @bittremieux.
made `n_missing_predictions` keyword-only so positional `score_bins` callers keep working & suppressed the redundant empty-report warning

casanovo/utils.py
```diff
@@ -206,9 +211,10 @@ def log_annotate_report(
     )
 
     if run_report is None:
-        logger.warning(
-            "No predictions were logged, this may be due to an error"
-        )
+        if n_missing_predictions == 0:
+            logger.warning(
+                "No predictions were logged, this may be due to an error"
```
bittremieux: **todo:** This warning message is incorrect. If `n_missing_predictions == 0`, that means that all spectra received a prediction. It can just be removed.

bittremieux (changes requested):

casanovo/utils.py
ssiddhantsharma: Removed the warning in 9521103 you're right, guarding it on `n_missing_predictions == 0` was contradictory.

ssiddhantsharma:
@bittremieux ready for re-review. Removed the incorrect `No predictions were logged` warning in 9521103 (updated its unit test). The missing-prediction count is reported via `log_annotate_report` (end-of-run report) as you requested, and the earlier CodeRabbit comments were resolved in 041d96b.

casanovo/utils.py
```diff
@@ -205,11 +210,7 @@ def log_annotate_report(
         score_bins=score_bins,
     )
 
-    if run_report is None:
```
bittremieux, line 208: **todo:** I'd still keep this original message in there, because explicitly warning the user that there are no predictions is useful. It was just that the previous check based on `n_missing_predictions` was incorrect.

bittremieux (changes requested):

casanovo/utils.py
ssiddhantsharma: Following your note, I kept the No predictions were logged warning and removed only the incorrect n_missing_predictions == 0 guard (https://github.com/Noble-Lab/casanovo/commit/0b60e6069ac07e2bd76c2c55d9d35cec31d45c76); the test now asserts the warning fires on an empty report. The missing-prediction count is reported via log_annotate_report, and the earlier CodeRabbit comments were resolved in https://github.com/Noble-Lab/casanovo/commit/041d96b695a98e23afce8dc1c6406be8b73e1baf.

bittremieux (approved):
Thanks!

PR #689, opened 2026-08-21
Report average precision and write precision-coverage TSV

### LINKED ISSUE #453: improve evaluate outputs

ssiddhantsharma:
Closes #453.

Evaluation now reports the average precision (area under the precision-coverage curve) alongside peptide and amino acid precision/recall, and writes the per-spectrum precision-coverage data to a TSV file next to the mzTab output (columns: `peak_file`, `scan_id`, `score`, `ground_truth_sequence`, `predicted_sequence`, `correct`, `precision`, `coverage`) so the precision-coverage curve can be plotted.

Added `precision_coverage` to `evaluate` with a unit test, and updated the `log_metrics` test to cover the new TSV output.

## Summary by CodeRabbit

* **New Features**
  * Added average precision and amino acid recall to evaluation metrics.
  * Added precision-coverage reporting for ranked peptide predictions.
  * Exported per-spectrum precision-coverage data as a TSV file alongside mzTab results.
  * Included evaluation fields—ground truth sequence, precision, and coverage—in mzTab output.
* **Documentation**
  * Documented the new evaluation metrics and TSV output in the Unreleased changelog.
* **Tests**
  * Added coverage for precision-coverage calculations, evaluation fields, and TSV generation.

### DIFF
--- a/casanovo/data/psm.py
+++ b/casanovo/data/psm.py
@@ -45,6 +45,10 @@ class PepSpecMatch:
     exp_mz: float
     aa_scores: Iterable[float]
     protein: str = "null"
+    # Evaluation results, populated only when sequencing with --evaluate.
+    ground_truth_sequence: Optional[str] = None
+    precision: Optional[float] = None
+    coverage: Optional[float] = None
 
     # Private properties to handle proteoform caching
     _proteoform_sequence: Optional[str] = dataclasses.field(
--- a/casanovo/data/psm.py
+++ b/casanovo/data/psm.py
@@ -46,9 +46,11 @@ class PepSpecMatch:
     aa_scores: Iterable[float]
     protein: str = "null"
     # Evaluation results, populated only when sequencing with --evaluate.
-    ground_truth_sequence: Optional[str] = None
-    precision: Optional[float] = None
-    coverage: Optional[float] = None
+    # The "null" default doubles as the mzTab empty-value token, so no
+    # None handling is needed when exporting these columns.
+    ground_truth_sequence: str = "null"
+    precision: float | str = "null"
+    coverage: float | str = "null"
 
     # Private properties to handle proteoform caching
     _proteoform_sequence: Optional[str] = dataclasses.field(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Added
 
 - The number of spectra that receive no prediction because beam search did not return a valid peptide is now counted and logged at the end of a sequencing run.
+- When sequencing with `--evaluate`, the ground truth sequence and the precision and coverage at each prediction are added as columns to the mzTab output, and the average precision (area under the precision-coverage curve) is reported in the end-of-run report.
 
 ### Changed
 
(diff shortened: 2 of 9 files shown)


bittremieux (commented):
Thanks for the contribution. If I understand the code changes correctly, it doesn't really implement the desired behavior unfortunately (the discussion in the issue was probably a bit underdefined).

What we want is the following: when running in evaluation mode (`casanovo sequence --evaluate`), output precision–coverage values in the mzTab.

During evaluation, the input is an annotated MGF, so we can calculate precision–coverage values. (In contrast during regular sequencing, there are no known peptide sequences specified in the MGF file.) In this case, we want to augment the mzTab results file that's exported at the end with some additional information:
- The true peptide sequence.
- The precision at this prediction.
- The coverage at this prediction.

These should be added as additional columns to the mzTab file with this information for each PSM. (Please look at the [zmTab documentation](https://raw.githubusercontent.com/HUPO-PSI/mzTab/master/specification_document-releases/1_0-Proteomics-Release/mzTab_format_specification.pdf) for column naming conventions.)

Additionally, we want to report the average precision in the end-of-run report.

What we do _not_ need to do is calculate precision–coverage during every validation step. In fact, this feature only applies to sequencing with evaluation, no change is needed during training.

ssiddhantsharma:
@bittremieux you're right, the issue asked for a separate TSV but embedding this in the mzTab is much cleaner. Rewritten in e13b7f3, thanks for the redirection, I am still figuring out the cases. 

The Precision–coverage is computed only when sequencing with `--evaluate` (no change during training/validation) and each PSM row gains three columns: `opt_global_ground_truth_sequence`, `opt_global_precision`, `opt_global_coverage`

Added a unit test asserting the attached precision/coverage/ground-truth values and an mzTab-writer test; the existing test_train_and_run covers the end-to-end evaluate path.

Average precision is reported in the end-of-run report & separate TSV is removed.
I used `opt_global_*` free-text names following the mzTab optional-column convention.

casanovo/data/psm.py
```diff
@@ -45,6 +45,10 @@ class PepSpecMatch:
     exp_mz: float
     aa_scores: Iterable[float]
     protein: str = "null"
+    # Evaluation results, populated only when sequencing with --evaluate.
+    ground_truth_sequence: Optional[str] = None
```
bittremieux: **suggestion:** You could have default value `"null"` for these instead of `None`, then there's always a value and you don't need the `None` check when exporting to mzTab in `ms_io`..

bittremieux (changes requested):
Looks good, I have just one minor comment. Can you also address the merge conflicts? Thanks.

casanovo/data/psm.py
ssiddhantsharma: Done in b35ebce defaulted ground_truth_sequence/precision/coverage to "null" (matching protein), so the None checks in ms_io are gone. The evaluate-only column gating now keys off precision != "null", so the columns still appear only when sequencing with --evaluate. Also rebased onto dev to clear the conflicts.

PR #693, opened 2026-08-26
Keep the learning rate within a half cosine period

josiehong:
In our experiments, keeping the learning rate on the decay for the whole run trained better, so the cosine half-period now spans the total number of training steps: the learning rate warms up, decays, and ends near zero on the final step, whatever `max_epochs` is set to.

```
cosine_schedule_period_iters = estimated_stepping_batches
                           ≈ max_epochs × ⌈batches_per_epoch / accumulate_grad_batches⌉
```

Changes:
- `Spec2Pep.configure_optimizers` sets the cosine half period from the trainer's `estimated_stepping_batches`, which accounts for epochs, gradient accumulation, and multi-device training; the init parameter is removed.
- The `cosine_schedule_period_iters` config option is removed. It and the old `max_iters` alias are registered in `_config_deprecated`.
- `config.yaml` and `config_timstof.yaml` document the derived schedule.
- FAQ updated (schedule description plus a note on `estimated_stepping_batches` for custom schedulers); the example mzTab header in the file formats docs no longer lists the removed option.
- CHANGELOG entries under Changed and Removed.

## Summary by CodeRabbit

- **Changed**
  - Learning-rate cosine decay now spans all training steps, begins decreasing after warm-up, and approaches zero at the final step.
  - The decay period is calculated automatically from the training run length.

- **Removed**
  - Removed the `cosine_schedule_period_iters` configuration option. Existing configurations and checkpoints containing it will show a deprecation warning, and the value will be ignored.

- **Documentation**
  - Updated configuration examples and FAQs to describe the new scheduling behavior and warm-up settings.


wsnoble:
Have you verified that this setup gives good results empirically?

josiehong:
> Have you verified that this setup gives good results empirically?

Ah, not yet. I'll train a new model now.

